PR #6, opened 2022-08-24
fix: pragma restrictions

re1ro:
* [x] Making required pragma to be `^0.8.0` so SDK can be used with any solidity project


milapsheth (approved):
None of these contracts are used directly by the gateway/deposit service in `cgp-solidity`, right?

re1ro:
@milapsheth the `Upgradable` is going to be used by deposit and gas services. 
But the pragma is fixed to `0.8.9` there, so CGP won't be affected by that change

PR #7, opened 2022-08-24
fix: executables and utils

re1ro:
* [x] AxelarForecallable: making the forecaller to be a message sender
* [x] AxelarGasService: adding the interface to the project
* [x] Readme: added install and docs


contracts/interfaces/IAxelarExecutable.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```
Was this intentional?

contracts/interfaces/IAxelarForecallable.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.9;
```
milapsheth, line 3: Same

milapsheth (approved):

contracts/interfaces/IAxelarExecutable.sol
re1ro: Yeah I noticed that and decided to fix the pragma in a separate PR #6 because of amount of files affected

contracts/interfaces/IAxelarForecallable.sol
re1ro: Yeah I noticed that and decided to fix the pragma in a separate PR #6 because of amount of files affected

re1ro: Reverting those lines for clarity

milapsheth (approved):

PR #8, opened 2022-08-25
feat: token linker with gas service

re1ro:
* [x] `TokenLinker` and `NFTLinker`: using gas service
* [x] Removing `dist` from git and adding `.npmignore`


contracts/token-linker/TokenLinkerBase.sol
```diff
@@ -0,0 +1,65 @@
...
+        string memory thisAddress = address(this).toString();
+        bytes memory payload = abi.encode(to, amount);
+        uint256 gasValue = _lockNative() ? msg.value - amount : msg.value;
+        if (gasValue > 0) {
+            gasService.payNativeGasForContractCall{ value: gasValue }(
+                address(this),
+                destinationChain,
+                thisAddress,
+                payload,
+                msg.sender
```
milapsheth, line 42: If an application contract makes use of this function, but the user pays for the gas to call the app contract, then we will only refund the app contract and not the user.

milapsheth (approved):

dist/ConstAddressDeployer.json
```diff
@@ -116,8 +116,8 @@
...
...00000000000000000000000000000000600052604160045260246000fd5b600082601f83011261038357600080fd5b813567ffffffffffffffff8082111561039e5761039e610343565b604051601f8301601f19908116603f011681019082821181831017156103c6576103c6610343565b816040528381528660208588010111156103df57600080fd5b836020870160208301376000602085830101528094505050505092915050565b6000806040838503121561041257600080fd5b823567ffffffffffffffff81111561042957600080fd5b61043585828601610372565b95602094909401359450505050565b60008083601f84011261045657600080fd5b50813567ffffffffffffffff81111561046e57600080fd5b60208301915083602082850101111561048657600080fd5b9250929050565b600080600080606085870312156104a357600080fd5b843567ffffffffffffffff8111156104ba57600080fd5b6104c687828801610444565b90955093505060208501356001600160a01b03811681146104e657600080fd5b9396929550929360400135925050565b6000806000806060858703121561050c57600080fd5b843567ffffffffffffffff8082111561052457600080fd5b61053088838901610372565b955060208701359450604087013591508082111561054d57600080fd5b5061055a87828801610444565b95989497509550505050565b818382376000910190815291905056fea2646970667358221220472226407ac2c04d886a2005c768eb7355d59b2f9ff02850f40c54dadc84694864736f6c63430008090033",
```
milapsheth, line 119: Also shouldn't be checking this into the repo itself

sammy1991106 (approved):

PR #10, opened 2022-08-31
fix: add checks to const address deployer

milapsheth:


contracts/ConstAddressDeployer.sol
```diff
@@ -3,34 +3,61 @@
...
-        (bool success, ) = deployedAddress_.call(init);
-        if (!success) revert FailedInit();
-        emit Deployed(keccak256(bytecode), salt, deployedAddress_);
+        if (init.length == uint256(0)) revert EmptyInit();
+
+        deployedAddress_ = _deploy(bytecode, msg.sender, salt);
+
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory returnData) = deployedAddress_.call(init);
+        if (!success || returnData.length != uint256(0)) revert FailedInit();
```
sammy1991106, line 54:
```suggestion
        if (!success) revert FailedInit();
```
checking the returnData's length makes no sense. It may likely return something we just don't know.

contracts/ConstAddressDeployer.sol
```diff
@@ -3,34 +3,61 @@
...
     ) external returns (address deployedAddress_) {
-        bytes32 newSalt = keccak256(abi.encode(msg.sender, salt));
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            deployedAddress_ := create2(0, add(bytecode, 32), mload(bytecode), newSalt)
-        }
-        (bool success, ) = deployedAddress_.call(init);
-        if (!success) revert FailedInit();
-        emit Deployed(keccak256(bytecode), salt, deployedAddress_);
+        if (init.length == uint256(0)) revert EmptyInit();
```
sammy1991106, line 48: I don't see why empty init should not be allowed. Init can be done in the fallback function.

contracts/ConstAddressDeployer.sol
```diff
@@ -3,34 +3,61 @@
...
+
+        deployedAddress_ = _deploy(bytecode, msg.sender, salt);
```
sammy1991106, line 50:
```suggestion
        deployedAddress_ = _deploy(bytecode, msg.sender, keccak256(abi.encode(sender, salt)));
```

contracts/ConstAddressDeployer.sol
```diff
@@ -52,4 +79,18 @@ contract ConstAddressDeployer {
             )
         );
     }
+
+    function _deploy(bytes memory bytecode, address sender, bytes32 salt) internal returns (address deployedAddress_) {
+        if (bytecode.length == 0) revert EmptyBytecode();
+
+        bytes32 newSalt = keccak256(abi.encode(sender, salt));
```
sammy1991106, line 86:
```suggestion
```

sammy1991106 (changes requested):

contracts/ConstAddressDeployer.sol
Foivos: we can remove the `sender` argument from _deploy and pass the hash as you suggest here but passing both the sender and the hash of the sender and the salt is redundant.

contracts/ConstAddressDeployer.sol
Foivos: If the contract can be initialised without any data then that should be done in the constructor instead. This check avoid user error.

Foivos (approved):

contracts/ConstAddressDeployer.sol
milapsheth: I was thinking that they could call the fallback explicitly if they wanted too. More concerned about user error.

contracts/ConstAddressDeployer.sol
milapsheth: Similar to a constructor, doesn't make sense to return something from the initializer, since we can't handle it. I'm ok with relaxing this though. Another option is to return it to the application from this function.

contracts/ConstAddressDeployer.sol
milapsheth: I would like to include the original `salt` in the emitted event below and reduce duplication.

contracts/ConstAddressDeployer.sol
sammy1991106: I don't see why we care if it returns or not. It doesn't matter in the context of a deployer. And constructor does not return in solidity. Our init method in gateway is `setup`, which does not return anything.

contracts/ConstAddressDeployer.sol
sammy1991106: you can't make a call to the fallback explicitly. Ya exactly if everything can be done in the constructor, why do we not allow the user to save some gas here by not doing anything? I don't see why are we dictating that.

contracts/ConstAddressDeployer.sol
sammy1991106: I didn't realize that sender is only used to calculate the salt down the line. Sure.

contracts/ConstAddressDeployer.sol
sammy1991106: but why? The original salt can be found in the tx data, which is even easier to be located than something in the event.

milapsheth: Easier to spot it in events in the explorer, but I'm ok with showing the derived salt too.

contracts/ConstAddressDeployer.sol
milapsheth: Ok, I removed the check, they can trigger the fallback through this if they want

contracts/ConstAddressDeployer.sol
milapsheth: Removed the check

Foivos (approved):

sammy1991106 (approved):

PR #20, opened 2023-02-10
feat(Express): new ExpressProxyDeployer, Create3Deployer and FinalProxy

re1ro:
- [x] `ExpressProxyDeployer`: express proxy factory
- [x] `Create3Deployer`: deploying to the same address based only on salt
- [x] `FinalProxy`: proxy that can deploy a final implementation


contracts/upgradable/FinalProxy.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IProxy } from '../interfaces/IProxy.sol';
+import { IFinalProxy } from '../interfaces/IFinalProxy.sol';
+import { Create3 } from '../deploy/Create3.sol';
+import { Proxy } from './Proxy.sol';
+
+contract FinalProxy is Proxy, IFinalProxy {
```
Foivos, line 10: Maybe add a comment explaining why this is better than just setting the `owner` to `address(0)` (aka it saves gas on every call after it is finalized)

Foivos (approved):

eguajardo (approved):

PR #24, opened 2023-02-22
fix(Create3): handling edge cases

re1ro:
* [x] `Create3`: handling errors self-destructing deployments
* [x] `Create3Deployer`: fixing comments

### DIFF
--- a/contracts/deploy/Create3Deployer.sol
+++ b/contracts/deploy/Create3Deployer.sol
@@ -38,7 +38,7 @@ contract Create3Deployer {
      * Requirements:
      *
      * - `bytecode` must not be empty.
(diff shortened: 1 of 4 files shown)


contracts/deploy/Create3Deployer.sol
```diff
@@ -38,7 +38,7 @@ contract Create3Deployer {
      * Requirements:
      *
      * - `bytecode` must not be empty.
-     * - `salt` must have not been used for `bytecode` already by the same `msg.sender`.
+     * - `salt` must have not been used already by the same `msg.sender`.
```
milapsheth, line 41:
```suggestion
     * - `salt` must not have been used already by the same `msg.sender`.
```
also applies to the previous method

milapsheth (approved):

PR #31, opened 2023-04-06
test: proxy test coverage

deanamiel:
Added test coverage for different proxy contracts corresponding to test cases on notion. Feel free to edit or suggest additional test cases.

### DIFF
--- a/contracts/upgradable/Proxy.sol
+++ b/contracts/upgradable/Proxy.sol
@@ -13,7 +13,6 @@ contract Proxy is BaseProxy {
         bytes memory setupParams
     ) {
         if (owner == address(0)) revert InvalidOwner();
-        if (implementation() != address(0)) revert AlreadyInitialized();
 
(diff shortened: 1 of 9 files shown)


re1ro (approved):

contracts/upgradable/Proxy.sol
```diff
@@ -13,7 +13,6 @@ contract Proxy is BaseProxy {
         bytes memory setupParams
     ) {
         if (owner == address(0)) revert InvalidOwner();
-        if (implementation() != address(0)) revert AlreadyInitialized();
```
Foivos, line 16: Why is this check removed?

deanamiel: It seems like this check would never fail because it is called in the constructor and implementation() would always return the zero address. Let me know if you can think of a case where it would fail.

sammy1991106 (approved):

PR #32, opened 2023-04-14
fix(Create3): checking for empty account with native ether balance

re1ro:
* [x] `Create3`: checking codehash for `keccak256('')`
* [x] `Upgradable`: better only proxy check
* [x] `Proxy` test: removed solidity warnings

### DIFF
--- a/contracts/deploy/Create3.sol
+++ b/contracts/deploy/Create3.sol
@@ -26,8 +26,11 @@ library Create3 {
 
     function deploy(bytes32 salt, bytes memory bytecode) internal returns (address deployed) {
         deployed = deployedAddress(salt, address(this));
+        bytes32 existingCodeHash = deployed.codehash;
 
-        if (deployed.codehash != bytes32(0)) revert AlreadyDeployed();
+        // https://eips.ethereum.org/EIPS/eip-1052
+        // keccak256('') == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
+        if (existingCodeHash != bytes32(0) && existingCodeHash != 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) revert AlreadyDeployed();
         if (bytecode.length == 0) revert EmptyBytecode();
 
         // CREATE2
--- a/contracts/deploy/Create3.sol
+++ b/contracts/deploy/Create3.sol
@@ -25,6 +25,7 @@ contract CreateDeployer {
 
 library Create3 {
     using ContractAddress for address;
+
     bytes32 internal constant DEPLOYER_BYTECODE_HASH = keccak256(type(CreateDeployer).creationCode);
 
     function deploy(bytes32 salt, bytes memory bytecode) internal returns (address deployed) {
--- a/contracts/deploy/Create3.sol
+++ b/contracts/deploy/Create3.sol
@@ -21,16 +21,28 @@ contract CreateDeployer {
     }
 }
 
+library AddressUtils {
+    function isContract(address _address) internal view returns (bool) {
+        bytes32 existingCodeHash = _address.codehash;
+
+        // https://eips.ethereum.org/EIPS/eip-1052
+        // keccak256('') == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
+        return
+            existingCodeHash != bytes32(0) &&
+            existingCodeHash != 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
+    }
+}
+
 library Create3 {
+    using AddressUtils for address;
     bytes32 internal constant DEPLOYER_BYTECODE_HASH = keccak256(type(CreateDeployer).creationCode);
 
     function deploy(bytes32 salt, bytes memory bytecode) internal returns (address deployed) {
         deployed = deployedAddress(salt, address(this));
-        bytes32 existingCodeHash = deployed.codehash;
 
         // https://eips.ethereum.org/EIPS/eip-1052
         // keccak256('') == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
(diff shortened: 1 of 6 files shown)


contracts/upgradable/Upgradable.sol
```diff
@@ -8,10 +8,15 @@ import { Ownable } from '../utils/Ownable.sol';
...
+    address internal immutable implementationAddress;
+
+    constructor() {
+        implementationAddress = address(this);
+    }
 
     modifier onlyProxy() {
         // Prevent setup from being called on the implementation
-        if (implementation() == address(0)) revert NotProxy();
+        if (address(this) == implementationAddress) revert NotProxy();
```
sammy1991106, line 19: Is there a unit test for this?

sammy1991106 (approved):

contracts/upgradable/Upgradable.sol
```diff
@@ -8,10 +8,15 @@ import { Ownable } from '../utils/Ownable.sol';
...
```
deanamiel, line 20: This check is a bit confusing as the implementationAddress is immutable and is set to address(this) in the constructor, is it that address(this) when called in the constructor is actually the address of the proxy?

re1ro: When the implementation is called though `delegatecall` by the proxy. `address(this)` becomes the address of the proxy

deanamiel: Makes sense, so actually the opposite of what I initially thought

deanamiel (approved):

contracts/deploy/Create3.sol
```diff
@@ -26,8 +26,11 @@ library Create3 {
 
     function deploy(bytes32 salt, bytes memory bytecode) internal returns (address deployed) {
         deployed = deployedAddress(salt, address(this));
+        bytes32 existingCodeHash = deployed.codehash;
 
-        if (deployed.codehash != bytes32(0)) revert AlreadyDeployed();
+        // https://eips.ethereum.org/EIPS/eip-1052
+        // keccak256('') == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
+        if (existingCodeHash != bytes32(0) && existingCodeHash != 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470) revert AlreadyDeployed();
```
milapsheth, line 33: Can you add a unit test for this? To make this more reusable, maybe add a library method `isContract` and use it for `address` type? We'll likely need this again.

milapsheth, line 33: This also needs to be handled during the sanity check at line 44. For e.g. a contract can `selfdestruct` and use it's own address to send the remaining funds, so it's codehash won't be 0. I'd also be open to remove the sanity check on line 44 altogether and rely on line 39 to check for failure.

contracts/upgradable/Upgradable.sol
deanamiel: I'll add this test

contracts/deploy/Create3.sol
deanamiel: I can add tests, were you thinking of adding it to an existing library or should I create a new one?

deanamiel: I can add the additional check to line 44

milapsheth: New one, since existing libraries don't seem relevant to this

milapsheth: Thanks

contracts/deploy/Create3.sol
```diff
@@ -41,7 +53,7 @@ library Create3 {
         deployer.deploy(bytecode);
 
         // checking for codehash instead of code length to support contracts that selfdestruct in constructor
-        if (deployed.codehash == bytes32(0)) revert DeployFailed();
+        if (!deployed.isContract()) revert DeployFailed();
```
milapsheth, line 56: Do we want to keep this sanity check?

contracts/deploy/Create3.sol
```diff
@@ -21,13 +21,28 @@ contract CreateDeployer {
     }
 }
 
+library AddressUtils {
```
re1ro, line 24: @deanamiel could you please move this into its own file under the `utils` folder? So we can also use it in other places as well. Let's call it `ContractAddress` to match the convention we have for other utils.

contracts/deploy/Create3.sol
```diff
@@ -21,13 +21,28 @@ contract CreateDeployer {
...
 library Create3 {
+    using AddressUtils for address;
     bytes32 internal constant DEPLOYER_BYTECODE_HASH = keccak256(type(CreateDeployer).creationCode);
 
     function deploy(bytes32 salt, bytes memory bytecode) internal returns (address deployed) {
         deployed = deployedAddress(salt, address(this));
 
-        if (deployed.codehash != bytes32(0)) revert AlreadyDeployed();
+        // https://eips.ethereum.org/EIPS/eip-1052
+        // keccak256('') == 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
```
re1ro, line 44: I think we can remove this comment as `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470` moved into the library and irrelevant in this context.

contracts/deploy/Create3.sol
re1ro: @deanamiel please remove this line as we already doing the success check here
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/32/files#diff-e8da8b3175680e091c0f1df471b0fda0ab20b884aa1dea000c4dd2caab59cdebL15-L17

deanamiel: Sounds good

contracts/deploy/Create3.sol
deanamiel: Good point

contracts/deploy/Create3.sol
```diff
@@ -22,12 +24,13 @@ contract CreateDeployer {
 }
 
 library Create3 {
+    using ContractAddress for address;
```
milapsheth, line 27:
```suggestion
    using ContractAddress for address;

```
newline for readability

milapsheth (approved):

PR #33, opened 2023-04-23
chore: update npm deps and remove ganache

milapsheth:
- Update npm deps
- Removed ganache dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@axelar-network/axelar-gmp-sdk-solidity",
-  "version": "3.3.0",
+  "version": "3.4.0",
   "description": "Solidity GMP SDK and utilities provided by Axelar for cross-chain development",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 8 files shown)


scripts/constAddressDeployer.js
Foivos, line 1: Did you test this on testnets? I remember having issues there with gas estimation in the past which is why I had the complex gas estimation code which is removed here.

milapsheth: gas estimation doesn't work well on some testnets. But using the ganache simulated value will also be incorrect since gas used can differ. I think the only real solution is letting the user pass in a gas limit themselves. I already updated the `info/env.json` files to specify some gas limit bound

.solhint.json
```diff
@@ -2,7 +2,18 @@
...
+    ],
+    "func-visibility": [
+      "warn",
+      {
+        "ignoreConstructors": true
+      }
+    ]
   }
-}
+}
```
sammy1991106, line 19:
```suggestion
}

```

package.json
```diff
@@ -32,4 +32,4 @@
   "engines": {
     "node": ">=16"
   }
-}
+}
```
sammy1991106, line 35:
```suggestion
}

```

sammy1991106 (approved):

deanamiel (approved):

PR #38, opened 2023-06-09
fix: don't use hardcoded gas constants

milapsheth:
- Some chains use different default gas for native transfers, so we can't make any assumptions
AXE-1115


contracts/utils/SafeTransfer.sol
```diff
@@ -39,14 +39,12 @@ library SafeTokenTransferFrom {
...
 library SafeNativeTransfer {
-    uint256 internal constant NATIVE_TRANSFER_GAS_LIMIT = 2300;
-
     function safeNativeTransfer(address receiver, uint256 amount) internal {
         bool success;
 
         // solhint-disable-next-line no-inline-assembly
         assembly {
-            success := call(NATIVE_TRANSFER_GAS_LIMIT, receiver, amount, 0, 0, 0, 0)
+            success := call(gasleft(), receiver, amount, 0, 0, 0, 0)
```
milapsheth, line 47:
```suggestion
            success := call(gas(), receiver, amount, 0, 0, 0, 0)
```

deanamiel (approved):

PR #39, opened 2023-06-16
feat: operators contract as per ticket axe680

deanamiel:
Contract described in ticket AXE-680. It is a pretty straightforward implementation of having multiple operators any of whom can call any arbitrary contract. There is also onlyOwner functionality to add and remove operators.

Note: This PR has been moved here from the cgp repo. Previous PR can be found here: https://github.com/axelarnetwork/axelar-cgp-solidity/pull/195

Solidity Test Coverage Report for Operators.sol:
100% Statements 12/12 | 100% Branches 18/18 | 100% Functions 6/6 | 100% Lines 17/17


contracts/interfaces/IOperators.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```
everywhere

contracts/interfaces/IOperators.sol
```diff
@@ -0,0 +1,22 @@
...
+
+interface IOperators {
```
milapsheth, line 5: add high level docs to the interface and keep implementation docs for implementation specific

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,93 @@
...
+import { Ownable } from './Ownable.sol';
+
+/**
+ * @title Operators
+ * @dev Contract module which provides an access control mechanism, where
+ * there is an account (an owner) that can be granted exclusive access to
+ * specific functions, and an operator that can execute arbitrary functions
+ * on any smart contract.
+ *
+ * The owner account is initially set as the deployer address.
```
milapsheth, line 15:
```suggestion
 * @dev This contract provides an access control mechanism, where
 * an owner can register operator accounts that can call arbitrary contracts on behalf of this contract.
 *
 * The owner account is initially set as the deployer address.
```
simplifying it a bit

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,93 @@
...
+    constructor() {
+        address initialOwner = msg.sender;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            sstore(_OWNER_SLOT, initialOwner)
+        }
+    }
+
+    /**
+     * @dev Modifier that requires the "msg.sender" to be an operator.
```
milapsheth, line 32:
```suggestion
     * @dev Modifier that requires the `msg.sender` to be an operator.
```

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,93 @@
...
+     * Emits a NotOperator error if the condition is not met.
```
milapsheth, line 33:
```suggestion
     * Reverts with a NotOperator error if the condition is not met.
```
perhaps, unless emit is also used in this context

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @dev Allows an operator to execute arbitrary functions on any smart contract.
+     * Can only be called by an operator.
+     * @param target address of the contract to execute the function on
+     * @param callData ABI encoded function call to execute on target
+     * @return data return data from executed function call
+     */
+    function execute(address target, bytes calldata callData) external onlyOperator returns (bytes memory) {
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory data) = target.call(callData);
```
milapsheth, line 86: forward `msg.value` sent by the operator so this is more generic?

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,93 @@
...
+
+        operators[operator] = false;
+
+        emit OperatorRemoved(operator);
+    }
+
+    /**
+     * @dev Allows an operator to execute arbitrary functions on any smart contract.
+     * Can only be called by an operator.
+     * @param target address of the contract to execute the function on
```
milapsheth, line 80:
```suggestion
     * @param target address of the contract to execute the function on. Existence is not checked.
```

contracts/interfaces/IOperators.sol
deanamiel: added

contracts/utils/Operators.sol
deanamiel: reverts with sounds better

contracts/utils/Operators.sol
deanamiel: Good idea

contracts/interfaces/IOperators.sol
```diff
@@ -1,7 +1,12 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.9;
+pragma solidity ^0.8.0;
 
+/**
+ * @title IOperators Interface
+ * @notice Interface for an access control mechanism with operators that can be given
+ * exlcusive permissions to execute functions.
```
milapsheth, line 8:
```suggestion
 * @notice Interface for an access control mechanism where operators have exclusive permissions to execute functions.
```

contracts/utils/Operators.sol
```diff
@@ -77,13 +74,12 @@ contract Operators is Ownable, IOperators {
...
      * Can only be called by an operator.
-     * @param target address of the contract to execute the function on
+     * @param target address of the contract to execute the function on. Existence is not checked.
      * @param callData ABI encoded function call to execute on target
      * @return data return data from executed function call
      */
-    function execute(address target, bytes calldata callData) external onlyOperator returns (bytes memory) {
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool success, bytes memory data) = target.call(callData);
+    function execute(address target, bytes calldata callData) external payable onlyOperator returns (bytes memory) {
```
milapsheth, line 81: based on a change we're adding to the express service, there's a need to allow configuring the msg value being forwarded
```suggestion
    function execute(address target, bytes calldata callData, unit256 nativeValue) external payable onlyOperator returns (bytes memory) {
```

contracts/utils/Operators.sol
```diff
@@ -77,13 +74,12 @@ contract Operators is Ownable, IOperators {
...
+        (bool success, bytes memory data) = target.call{ value: msg.value }(callData);
```
milapsheth, line 82:
```suggestion
        (bool success, bytes memory data) = target.call{ value: nativeValue }(callData);
```
for convenience, could also make it so that msg.value is used if nativeValue is `0`. what do you think?

contracts/utils/Operators.sol
deanamiel: Changed this in the interface as well

contracts/utils/Operators.sol
deanamiel: If configuring the nativeValue in the function call is optional then yeah, I think we should account for a nativeValue of zero but a nonzero msg.value.

contracts/utils/Operators.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        emit OperatorRemoved(operator);
+    }
+
+    /**
+     * @dev Allows an operator to execute arbitrary functions on any smart contract.
+     * Can only be called by an operator.
+     * @param target address of the contract to execute the function on. Existence is not checked.
+     * @param callData ABI encoded function call to execute on target
+     * @param nativeValue The amount of native asset to send with the call
```
milapsheth, line 79:
```suggestion
     * @param nativeValue The amount of native asset to send with the call. If `nativeValue` is set to `0`, then `msg.value` is forwarded instead.
```

contracts/test/TestOperators.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.9;
```
milapsheth, line 3:
```suggestion
pragma solidity ^0.8.0;
```

milapsheth (approved):

PR #48, opened 2023-07-12
feat: add slither config

milapsheth:


contracts/utils/AddressString.sol
```diff
@@ -36,7 +36,7 @@ library AddressToString {
         stringBytes[0] = '0';
         stringBytes[1] = 'x';
 
-        for (uint256 i; i < length; ++i) {
+        for (uint256 i = 0; i < length; ++i) {
```
Foivos, line 39: In a previous code arena they told us this is less efficient, and they are equivalent otherwise. I haven't tested it tho.

Foivos (approved):

contracts/utils/AddressString.sol
milapsheth: ok, slither warning, but will revert.

milapsheth, line 39:
```suggestion
        for (uint256 i; i < length; ++i) {
```

deanamiel (approved):

PR #49, opened 2023-07-18
fear: using arbitrary gasOptions instead of only gasLimit for create2 and create3

Foivos:


scripts/constAddressDeployer.js
```diff
@@ -46,7 +46,7 @@ const deployContractConstant = async (
   contractJson,
   key,
   args = [],
-  gasLimit = null,
```
milapsheth, line 49: there are external dependencies on this. can you add txOptions as a separate param and use that if gasLimit isn't provided?

Foivos: Instead of a separate param we check if it is a number and treat it as the gasLimit, otherwise as the full options

milapsheth (approved):

PR #51, opened 2023-07-25
feat: add receive to operators

deanamiel:
PR to add a receive method to the Operators contract.

### DIFF
--- a/test/Operators.js
+++ b/test/Operators.js
@@ -112,6 +112,21 @@ describe('Operators', () => {
         operators.connect(ownerWallet).removeOperator(userWallet.address),
       ).to.be.revertedWithCustomError(operators, 'NotAnOperator');
     });
+
+    it('should receive ether', async () => {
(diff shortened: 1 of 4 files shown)


test/Operators.js
```diff
@@ -112,6 +112,21 @@ describe('Operators', () => {
...
+    it('should receive ether', async () => {
+      const sendValue = 100;
+
+      const initBalance = await ethers.provider.getBalance(operators.address);
+      expect(initBalance).to.equal(0);
+
+      await ownerWallet.sendTransaction({
+        to: operators.address,
+        value: sendValue,
+      });
```
milapsheth, line 125:
```suggestion
      await ownerWallet.sendTransaction({
        to: operators.address,
        value: sendValue,
      }).then((tx) => tx.wait());
```
always wait on the tx confirmation

milapsheth (approved):

PR #54, opened 2023-08-01
fix: initial owner as constructor arg

deanamiel:
PR to pass the initial owner of Operators as a constructor arg.

### DIFF
--- a/test/Operators.js
+++ b/test/Operators.js
@@ -3,6 +3,8 @@
 const chai = require('chai');
 const { ethers } = require('hardhat');
 const { expect } = chai;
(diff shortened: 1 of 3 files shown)


milapsheth (approved):

milapsheth:
bump version number

test/Operators.js
```diff
@@ -219,4 +237,44 @@ describe('Operators', () => {
       expect(targetBalance).to.equal(msgValue);
     });
   });
+
+  describe('custom deployer', () => {
```
milapsheth, line 241: deployment tests should be moved to contract-deployments repo later

PR #55, opened 2023-08-07
feat: deployer interfaces

deanamiel:
PR with some updates to the deployer files and new interfaces for create2/create3 deployers.

Note: ConstAddressDeployer does not inherit its interface because that would result in bytecode differences with existing deployments.


re1ro (approved):

contracts/interfaces/IConstAddressDeployer.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title IConstAddressDeployer Interface
+ * @notice This interface defines the contract responsible for deploying and optionally initializing new contracts
+ * using the `CREATE2` method.
+ */
+// slither-disable-next-line missing-inheritance
```
milapsheth, line 10: this can probably be removed after inheriting in Create2Deployer

contracts/interfaces/IConstAddressDeployer.sol
```diff
@@ -0,0 +1,51 @@
...
+interface IConstAddressDeployer {
```
milapsheth, line 11:
```suggestion
interface ICreate2Deployer {
```
since we're renaming the deployer

milapsheth (approved):

contracts/interfaces/IConstAddressDeployer.sol
deanamiel: removed

contracts/interfaces/IConstAddressDeployer.sol
deanamiel: changed the name

PR #59, opened 2023-08-11
feat(Create3Deployer): passing native value for deployment

re1ro:
* [x] `Create3Deployer`: passing native value for deployment

### DIFF
--- a/test/Create3Deployer.js
+++ b/test/Create3Deployer.js
@@ -154,7 +154,7 @@ describe('Create3Deployer', () => {
       expect(await contract.decimals()).to.equal(decimals);
     });
 
-    it('should deploy with passing native value to the contact', async () => {
+    it('should deploy with native value to the contract', async () => {
       const key = 'a test key';
       // Send 1 eth to address
       const amount = ethers.utils.parseEther('1');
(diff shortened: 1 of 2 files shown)


test/Create3Deployer.js
```diff
@@ -154,6 +154,25 @@ describe('Create3Deployer', () => {
       expect(await contract.decimals()).to.equal(decimals);
     });
 
+    it('should deploy with passing native value to the contact', async () => {
```
deanamiel, line 157: Should we add test for deployAndInit with native value?

deanamiel (approved):

test/Create3Deployer.js
re1ro: Yeah that's a good idea

test/Create3Deployer.js
```diff
@@ -154,6 +154,25 @@ describe('Create3Deployer', () => {
...
+      const key = 'a test key';
+      // Send 1 eth to address
+      const amount = ethers.utils.parseEther('1');
```
milapsheth, line 160:
```suggestion
      // Send eth to address
      const amount = ethers.utils.parseEther('0.00000001');
```
This is expensive when running on a live network

milapsheth (approved):

PR #60, opened 2023-08-11
fix(Proxy): checking contractId for final and fixed proxies

re1ro:
* [x] `FinalProxy` and `FixedProxy`: checking `contractId()` for final and fixed proxies

### DIFF
--- a/contracts/interfaces/IContractIdentifier.sol
+++ b/contracts/interfaces/IContractIdentifier.sol
@@ -5,8 +5,8 @@ pragma solidity ^0.8.0;
 // General interface for upgradable contracts
 interface IContractIdentifier {
     /**
-     * @dev Returns the contract ID. It can be used as a check during upgrades.
-     * @notice Meant to be overridden in derived contracts.
+     * @notice Returns the contract ID. It can be used as a check during upgrades.
+     * @dev Meant to be overridden in derived contracts.
      * @return bytes32 The contract ID
      */
     function contractId() external pure returns (bytes32);
--- a/contracts/interfaces/IContractIdentifier.sol
+++ b/contracts/interfaces/IContractIdentifier.sol
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+// General interface for upgradable contracts
+interface IContractIdentifier {
+    /**
+     * @dev Returns the contract ID. It can be used as a check during upgrades.
+     * @notice Meant to be overridden in derived contracts.
+     * @return bytes32 The contract ID
+     */
+    function contractId() external pure returns (bytes32);
+}
--- a/contracts/upgradable/BaseProxy.sol
+++ b/contracts/upgradable/BaseProxy.sol
@@ -33,7 +33,7 @@ abstract contract BaseProxy is IProxy {
 
     /**
      * @dev Returns the contract ID. It can be used as a check during upgrades.
-     * Meant to be overridden in derived contracts.
+     * @notice Meant to be overridden in derived contracts.
      * @return bytes32 The contract ID
      */
     function contractId() internal pure virtual returns (bytes32) {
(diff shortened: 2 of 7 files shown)


contracts/upgradable/BaseProxy.sol
```diff
@@ -33,7 +33,7 @@ abstract contract BaseProxy is IProxy {
 
     /**
      * @dev Returns the contract ID. It can be used as a check during upgrades.
-     * Meant to be overridden in derived contracts.
+     * @notice Meant to be overridden in derived contracts.
      * @return bytes32 The contract ID
```
deanamiel, line 37: Switch notice and dev tags?

contracts/upgradable/FinalProxy.sol
```diff
@@ -79,7 +80,11 @@ contract FinalProxy is Proxy, IFinalProxy {
         }
         if (msg.sender != owner) revert NotOwner();
 
+        bytes32 id = contractId();
         finalImplementation_ = Create3.deploy(FINAL_IMPLEMENTATION_SALT, bytecode);
+
+        if (id != bytes32(0) && IUpgradable(finalImplementation_).contractId() != id) revert InvalidImplementation();
+
```
deanamiel, line 87: Should we add unit test for this and for fixed proxy?

deanamiel (approved):

contracts/upgradable/FinalProxy.sol
re1ro: Yes I think so. We will have to eventually anyway to get 100% test coverage again 😄

contracts/upgradable/BaseProxy.sol
re1ro: Yeah makes sense

contracts/upgradable/FinalProxy.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.0;
 
 import { IProxy } from '../interfaces/IProxy.sol';
 import { IFinalProxy } from '../interfaces/IFinalProxy.sol';
+import { IUpgradable } from '../interfaces/IUpgradable.sol';
```
milapsheth, line 7: Can we split contractId() into a separate interface (maybe `IContractIdentifier` or so)? Atm, we're pulling in the full IUpgradable which can be an issue for the bytecode if upgradable contract is changed

contracts/upgradable/FinalProxy.sol
```diff
@@ -79,7 +80,11 @@ contract FinalProxy is Proxy, IFinalProxy {
         }
         if (msg.sender != owner) revert NotOwner();
 
+        bytes32 id = contractId();
         finalImplementation_ = Create3.deploy(FINAL_IMPLEMENTATION_SALT, bytecode);
+
+        if (id != bytes32(0) && IUpgradable(finalImplementation_).contractId() != id) revert InvalidImplementation();
```
milapsheth, line 86: Needs a comment around why the exception for the empty case

contracts/upgradable/FinalProxy.sol
re1ro: The `IUpgradable`  interface most likely will stay the same event we we modify the `Upgradable`. Plus with Create3 we should not worried about the bytecode anymore.

contracts/upgradable/FinalProxy.sol
re1ro: Added

contracts/upgradable/FinalProxy.sol
re1ro: Added `IContractIdentifier` and made other contracts depend on it

contracts/interfaces/IContractIdentifier.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+// General interface for upgradable contracts
+interface IContractIdentifier {
+    /**
+     * @dev Returns the contract ID. It can be used as a check during upgrades.
+     * @notice Meant to be overridden in derived contracts.
```
milapsheth, line 9:
```suggestion
     * @notice Returns the contract ID. It can be used as a check during upgrades.
     * @dev Meant to be overridden in derived contracts.
```

milapsheth (approved):

PR #63, opened 2023-08-11
feat(ownable): ownable updates

deanamiel:
PR to add a constructor to Ownable.sol


contracts/utils/Ownable.sol
```diff
@@ -94,4 +93,20 @@ abstract contract Ownable is IOwnable {
...
         }
     }
+
+    /**
+     * @notice Internal function to transfer ownership of the contract to a new account `newOwner`.
+     * @dev Called in the constructor to set the initial owner.
+     * @param newOwner The address to transfer ownership to
+     */
+    function _transferOwnership(address newOwner) internal virtual {
+        if (newOwner == address(0)) revert ZeroAddress();
```
milapsheth, line 103:
```suggestion
        if (newOwner == address(0)) revert InvalidAddress();
```
Perhaps?

milapsheth (approved):

re1ro (approved):

contracts/utils/Ownable.sol
re1ro: Yeah also needs to be renamed in the interface

deanamiel: Changed to `InvalidOwnerAddress `

PR #69, opened 2023-08-16
fix(Upgradable): remove owner constructor arg

deanamiel:
PR to remove the owner constructor arg from Upgradable.sol


contracts/upgradable/Upgradable.sol
```diff
@@ -17,9 +17,11 @@ abstract contract Upgradable is Ownable, IUpgradable {
     /**
      * @notice Constructor sets the implementation address to the address of the contract itself
      * @dev This is used in the onlyProxy modifier to prevent certain functions from being called directly
-     * on the implementation contract itself
+     * on the implementation contract itself.
+     * @dev The owner is initially set as address(1) because the actual owner is set within the proxy's init function. It is not
```
milapsheth, line 21:
```suggestion
     * @dev The owner is initially set as address(1) because the actual owner is set within the proxy. It is not
```
might not be in init depending on the proxy

milapsheth (approved):

PR #71, opened 2023-08-21
fix: improve various checks and tests

milapsheth:
- [x] Reuse methods in Ownable contract
- [x] Refactor Ownable unit test
- [x] Optimize String/Address conversion
- [x] Remove redundant express executable check
- [x] Move default `contractId` implementation


contracts/express/AxelarValuedExpressExecutable.sol
```diff
@@ -225,7 +225,7 @@ contract AxelarValuedExpressExecutable is ExpressExecutorTracker {
 
         if (tokenAddress == address(0)) {
             if (value != msg.value) revert InsufficientValue();
-        } else if (value > 0) {
```
milapsheth, line 228: value == 0 is already checked at the top of the method

contracts/upgradable/BaseProxy.sol
```diff
@@ -32,13 +32,10 @@ abstract contract BaseProxy is IProxy {
...
     /**
-     * @dev Returns the contract ID. It can be used as a check during upgrades.
-     * @notice Meant to be overridden in derived contracts.
+     * @dev Returns the contract ID. It can be used as a check during upgrades. Meant to be implemented in derived contracts.
      * @return bytes32 The contract ID
      */
-    function contractId() internal pure virtual returns (bytes32) {
-        return bytes32(0);
-    }
+    function contractId() internal pure virtual returns (bytes32);
```
milapsheth, line 38: Moved the default implementation to the Proxy that expects that behaviour

deanamiel (approved):

re1ro (approved):

PR #81, opened 2023-08-30
feat(IAxelarGasService): natspec docs

deanamiel:
PR for IAxelarGasService NatSpec documentation.


contracts/interfaces/IAxelarGasService.sol
```diff
@@ -216,45 +244,100 @@ interface IAxelarGasService is IUpgradable {
...
+     * @dev Use address(0) as the token address to refund native currency.
+     * @param txHash The transaction hash of the cross-chain call
+     * @param logIndex The log index for the cross-chain call
+     * @param receiver The address to receive the refund
+     * @param token The token address to be refunded
+     * @param amount The amount to refund
+     */
     function refund(
+        bytes32 txHash,
+        uint256 logIndex,
```
deanamiel, line 332: I changed this to the function definition of the non-deprecated `refund` function, not sure if it was kept as the deprecated function definition intentionally for backward compatibility or something

milapsheth: ok, should be fine

milapsheth (approved):

PR #97, opened 2023-10-11
fix(Deployer): passing native value in the Deployer instead of Create2/3 lib

re1ro:
* [x] `Deployer`: passing native value in the `Deployer` instead of `Create2` and `Create3` libs


deanamiel (approved):

contracts/test/deploy/TestCreste2.sol
milapsheth, line 1: Should be TestCreate2

contracts/test/deploy/TestCreste3.sol
milapsheth, line 1: Should be TestCreate3

contracts/test/deploy/TestCreate2.sol
```diff
@@ -0,0 +1,15 @@
+import { Create2 } from '../../deploy/Create2.sol';
```
milapsheth, line 1:
```suggestion
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.0;

import { Create2 } from '../../deploy/Create2.sol';
```

contracts/test/deploy/TestCreate3.sol
```diff
@@ -0,0 +1,15 @@
+import { Create3 } from '../../deploy/Create3.sol';
```
milapsheth, line 1:
```suggestion
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.0;

import { Create3 } from '../../deploy/Create3.sol';
```

milapsheth (approved):

PR #99, opened 2023-10-13
feat(Roles): new Roles contract + tests

re1ro:
[AXE-2202]
* [x] `BaseRoles`: utility contract with internal functions that allows accounts to have many roles.
* [x] `Roles`: implementation with included set of functions for transferring roles.

[AXE-2202]: https://axelarnetwork.atlassian.net/browse/AXE-2202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/utils/Roles.sol
```diff
@@ -0,0 +1,359 @@
...
+
+        _transferRoles(fromAccount, msg.sender, roles);
+    }
+
+    /**
+     * @notice Internal function to convert an array of roles to a uint256.
+     * @param roles The roles to convert
+     * @return accountRoles The roles in uint256 format
+     */
+    function _toAccountRoles(uint8[] calldata roles) internal returns (uint256) {
```
deanamiel, line 176: Change state mutability to `pure`?

contracts/utils/Roles.sol
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /**
+     * @notice Transfers roles of the contract to a new account.
+     * @dev Can only be called by the account with all the roles.
+     * @dev Emits RolesRemoved and RolesAdded events.
+     * @param toAccount The address to transfer role to
+     * @param roles The roles to transfer
+     */
+    function transferRoles(address toAccount, uint8[] calldata roles) external virtual withAllTheRoles(roles) {
```
deanamiel, line 130: Do we want to add the same sanity checks as `proposeRoles`?
`if (toAccount == address(0) || toAccount == msg.sender) revert InvalidTransferAccount(toAccount);`

contracts/utils/Roles.sol
```diff
@@ -0,0 +1,359 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IRoles } from '../interfaces/IRoles.sol';
+
+/**
+ * @title Roles
+ * @notice A contract module which provides a basic access control mechanism, where
+ * there is an account (an role) that can be granted exclusive access to
```
deanamiel, line 10:
```suggestion
 * there is an account (a role) that can be granted exclusive access to
```

contracts/interfaces/IRoles.sol
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @notice Returns the roles of an account.
+     * @param account The address to get the roles for
+     * @return accountRoles The roles of the account in uint256 format
+     */
+    function getAccountRoles(address account) external view returns (uint256 accountRoles);
+
+    /**
+     * @notice Returns the pending role of the contract.
+     * @param fromAccount The address of the current role
```
deanamiel, line 58:
```suggestion
     * @param fromAccount The address with the current roles
```

contracts/interfaces/IRoles.sol
```diff
@@ -0,0 +1,91 @@
...
+     * @param toAccount The address of the pending role
```
deanamiel, line 59:
```suggestion
     * @param toAccount The address with the pending roles
```
If you like this wording better I can update it for all the other comments, let me know

contracts/utils/Roles.sol
```diff
@@ -0,0 +1,359 @@
...
+            revert MissingAllRoles(msg.sender, roles);
+        }
+
+        _;
+    }
+
+    /**
+     * @notice Modifier that throws an error if called by an account without any of the roles.
+     */
+    modifier withAnyOfRoles(uint8[] calldata roles) {
```
Foivos, line 66: this should be memory so that constants can be passed in.

contracts/utils/Roles.sol
Foivos: also change from calldata to memory

contracts/utils/Roles.sol
```diff
@@ -0,0 +1,281 @@
...
+        return true;
+    }
+
+    /**
+     * @notice Internal function to check if an account has any of the roles.
+     * @param accountRoles The roles of the account in uint256 format
+     * @param roles The roles to check
+     * @return True if the account has any of the roles, false otherwise
+     */
+    function _hasAnyOfRoles(uint256 accountRoles, uint8[] memory roles) internal pure returns (bool) {
```
milapsheth, line 220: don't think we need to add this due to bytecode size increase

contracts/utils/RoleTransfer.sol
```diff
@@ -0,0 +1,106 @@
...
+ * change until the pending role accepts the role transfer.
+ */
+contract RoleTransfer is Roles, IRoleTransfer {
+    /**
+     * @notice Checks if an account has a role.
+     * @param account The address to check
+     * @param role The role to check
+     * @return True if the account has the role, false otherwise
+     */
+    function hasRole(address account, uint8 role) public view returns (bool) {
```
milapsheth, line 26: can you move this to the base role contract?

contracts/utils/Roles.sol
re1ro: It's not included in the bytecode if inheriting contract is not using it.

deanamiel (approved):

milapsheth (approved):

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * @title RolesBase
+ * @notice A contract module which provides a set if internal functions
+ * for implementing role control features.
+ */
+contract RolesBase is IRolesBase {
+    bytes32 internal constant ROLES_PREFIX = keccak256('roles');
+    bytes32 internal constant ROLES_PROPOSAL_PREFIX =
+       keccak256('roles-proposal');
```
milapsheth, line 15:
```suggestion
    bytes32 internal constant ROLES_PROPOSAL_PREFIX = keccak256('roles-proposal');
```

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+     * @dev Can only be called by the account with all the proposed roles.
+     * @dev emits a RolesProposed event.
+     * @dev Roles are not transferred until the new role accepts the role transfer.
+     * @param toAccount The address to transfer role to
+     * @param roles The roles to transfer
+     */
+    function _proposeRoles(address toAccount, uint8[] memory roles) internal {
+        uint256 proposed = _toAccountRoles(roles);
+
+        _setProposedRoles(msg.sender, toAccount, proposed);
```
milapsheth, line 244: don't see the check that msg.sender has the `proposed` roles already to allow proposing it for another account

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /**
+     * @notice Internal function to get the key of the proposed roles mapping.
+     * @param fromAccount The address of the current role
+     * @param toAccount The address of the pending role
+     * @return key The key of the proposed roles mapping
+     */
+    function _proposalKey(address fromAccount, address toAccount) internal view virtual returns (bytes32 key) {
+        return keccak256(abi.encodePacked(ROLES_PROPOSAL_PREFIX, fromAccount, toAccount));
```
milapsheth, line 115:
```suggestion
        return keccak256(abi.encodePacked(PROPOSE_ROLES_PREFIX, fromAccount, toAccount));
```

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+import { IRolesBase } from '../interfaces/IRolesBase.sol';
+
+/**
+ * @title RolesBase
+ * @notice A contract module which provides a set if internal functions
+ * for implementing role control features.
+ */
+contract RolesBase is IRolesBase {
+    bytes32 internal constant ROLES_PREFIX = keccak256('roles');
+    bytes32 internal constant ROLES_PROPOSAL_PREFIX = keccak256('roles-proposal');
```
milapsheth, line 14:
```suggestion
    bytes32 internal constant PROPOSE_ROLES_PREFIX = keccak256('propose-roles');
```
these two share a common prefix string, should change

milapsheth (changes requested):

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * @notice Internal function to add roles to an account.
+     * @dev emits a RolesAdded event.
+     * @dev Called in the constructor to set the initial roles.
+     * @param account The address to add roles to
+     * @param roles The roles to add
+     */
+    function _addRoles(address account, uint8[] memory roles) internal {
+        uint256 accountRoles = _getRoles(account);
```
Foivos, line 156: This is a pain to work with, because it requires:
```
uint8[] memory roles = new uint8[](1);
roles[0] = DISTRIBUTOR;
```
Maybe we can also add a function called _addRoles(address, uint8) that makes this easier for adding a single role.

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+            accountRoles := sload(key)
+        }
+    }
+
+    /**
+     * @notice Internal function to set the roles of an account.
+     * @param account The address to set the roles for
+     * @param accountRoles The roles to set
+     */
+    function _setRoles(address account, uint256 accountRoles) private {
```
Foivos, line 101: Why is this private?

contracts/utils/RolesBase.sol
re1ro: Yup will add functions `_addRole` and `_removeRole`

contracts/utils/RolesBase.sol
re1ro: it's implemented in the external function.

contracts/utils/RolesBase.sol
```diff
@@ -0,0 +1,282 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @notice Internal function to propose to transfer roles of message sender to a new account.
+     * @dev Can only be called by the account with all the proposed roles.
```
milapsheth, line 235:
```suggestion
     * @dev Caller MUST verify that msg.sender has all the proposed roles.
```

contracts/utils/RolesBase.sol
milapsheth: ok, doc is misleading,
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/99/files#r1362575370
Can you update it for other internal functions as well?

contracts/utils/RolesBase.sol
re1ro: The idea here is not to write to the storage directly bypassing event emission.

contracts/utils/RolesBase.sol
re1ro: Maybe we actually need to do a check inside of that internal function, since transfer roles always implies fromAccount should have those roles.

contracts/utils/RolesBase.sol
re1ro: Yup will add `_addRole` and `_removeRole` functions for convenience.

re1ro: @Foivos here is how you can actually do it. Wrapping it into a tuple
```
_addRoles(account, new uint8[]((DISTRIBUTOR)));
```

contracts/utils/RolesBase.sol
```diff
@@ -150,7 +150,9 @@ contract RolesBase is IRolesBase {
         accountRoles |= (1 << role);
         _setRoles(account, accountRoles);
 
-        emit RolesAdded(account, new uint8[]((role)));
+        uint8[] memory roles = new uint8[](1);
+        roles[0] = role;
+        emit RolesAdded(account, roles);
```
milapsheth, line 155: just call _addRoles instead since it's basically the same?

milapsheth (approved):

PR #107, opened 2023-10-26
refactor: add caller address in role related events [AXE-2407]

blockchainguyy:
Added caller address to be also emitted when a role is being assigned or removed for a account [AXE-2407](https://axelarnetwork.atlassian.net/browse/AXE-2407)

[AXE-2407]: https://axelarnetwork.atlassian.net/browse/AXE-2407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interfaces/IRolesBase.sol
+++ b/contracts/interfaces/IRolesBase.sol
@@ -15,8 +15,8 @@ interface IRolesBase {
     error InvalidProposedRoles(address fromAccount, address toAccount, uint8[] roles);
 
     event RolesProposed(address indexed fromAccount, address indexed toAccount, uint8[] roles);
-    event RolesAdded(address indexed account, uint8[] roles);
-    event RolesRemoved(address indexed account, uint8[] roles);
+    event RolesAdded(address indexed caller, address indexed account, uint8[] roles);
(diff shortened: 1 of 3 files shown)


contracts/interfaces/IRolesBase.sol
```diff
@@ -15,8 +15,8 @@ interface IRolesBase {
     error InvalidProposedRoles(address fromAccount, address toAccount, uint8[] roles);
 
     event RolesProposed(address indexed fromAccount, address indexed toAccount, uint8[] roles);
-    event RolesAdded(address indexed account, uint8[] roles);
-    event RolesRemoved(address indexed account, uint8[] roles);
+    event RolesAdded(address indexed caller, address indexed account, uint8[] roles);
```
milapsheth, line 18: What do we need this for? The way _addRoles etc. would be implemented usually would be to use msg.sender as the account itself. And in some instances, msg.sender can be ITS contract. The pattern of adding the caller to events feels weird since it could be applied to all events.

PR #110, opened 2023-10-31
feat: added some more utility functions for roles and made a few optimizations

Foivos:


contracts/utils/RolesBase.sol
```diff
@@ -281,12 +323,42 @@ contract RolesBase is IRolesBase {
...
+    function _acceptAccountRoles(
+        address fromAccount,
+        address toAccount,
+        uint256 accountRoles
+    ) internal virtual {
+        if (_getProposedRoles(fromAccount, toAccount) != accountRoles) {
+            revert InvalidProposedRoles(fromAccount, toAccount, accountRoles);
         }
 
         _setProposedRoles(fromAccount, toAccount, 0);
```
milapsheth, line 345: Shall we set proposed roles to 0 in transferRoles instead? Removes the edge case of having stale proposals if someone transfers directly

milapsheth (approved):

contracts/utils/RolesBase.sol
Foivos: i don't like the fact that transfers would now cost a bit more if they are not through a proposal (to set proposed roles from 0 to 0). There is also the edge case where I propose you become the OPERATOR and transfer the DISTRIBUTOR role to you after, causing the proposal of OPERATOR to be lost in that case, which would be a bug.

PR #112, opened 2023-11-01
fix(InterchainAddressTracker): removing Upgradable

re1ro:
* [x] `InterchainAddressTracker`: removing Upgradable dependency for composability


contracts/utils/InterchainAddressTracker.sol
```diff
@@ -29,14 +26,7 @@ contract InterchainAddressTracker is IInterchainAddressTracker, Upgradable {
...
 
-    /**
-     * @notice Getter for the contract id.
-     */
-    function contractId() external pure returns (bytes32) {
-        return CONTRACT_ID;
-    }
-
-    function _setup(bytes calldata params) internal override {
+    function _setupTrustedAddresses(bytes calldata params) internal {
```
milapsheth, line 29: Hmm, this should be handled by the inheriting contract, in the constructor?

contracts/test/utils/TestInterchainAddressTracker.sol
```diff
@@ -2,15 +2,32 @@
 
 pragma solidity ^0.8.0;
 
+import { Ownable } from "../../utils/Ownable.sol";
```
milapsheth, line 5:
```suggestion
import { Ownable } from '../../utils/Ownable.sol';
```

contracts/utils/InterchainAddressTracker.sol
re1ro: Oops meant to delete it

milapsheth (approved):

PR #116, opened 2023-11-17
fix(Upgradable): improved contract id verification

re1ro:
* [x] `Upgradable`: checking contract id against current implementation when upgrading
* [x] `InterchainAddressTracker`: prettier

### DIFF
--- a/contracts/upgradable/Implementation.sol
+++ b/contracts/upgradable/Implementation.sol
@@ -10,14 +10,10 @@ import { IImplementation } from '../interfaces/IImplementation.sol';
  * @dev Derived contracts must implement the setup function.
  */
 abstract contract Implementation is IImplementation {
-    address private immutable implementationAddress;
-
     /**
-     * @dev Contract constructor that sets the implementation address to the address of this contract.
+     * @dev Sets the implementation address to the address of this contract.
      */
-    constructor() {
-        implementationAddress = address(this);
-    }
+    address private immutable implementationAddress = address(this);
 
     /**
      * @dev Modifier to require the caller to be the proxy contract.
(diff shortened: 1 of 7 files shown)


contracts/upgradable/Implementation.sol
```diff
@@ -10,14 +10,10 @@ import { IImplementation } from '../interfaces/IImplementation.sol';
...
-    address private immutable implementationAddress;
-
     /**
-     * @dev Contract constructor that sets the implementation address to the address of this contract.
+     * @dev Sets the implementation address to the address of this contract.
      */
-    constructor() {
-        implementationAddress = address(this);
-    }
+    address private immutable implementationAddress = address(this);
```
deanamiel, line 16: Can this not be declared as a `constant` variable?

deanamiel (approved):

contracts/upgradable/Implementation.sol
re1ro: Constant is defined at the compilation time and deployed address would be unknown, immutable is defined at the deployment time

contracts/upgradable/Implementation.sol
```diff
@@ -10,14 +10,10 @@ import { IImplementation } from '../interfaces/IImplementation.sol';
  * @dev Derived contracts must implement the setup function.
  */
 abstract contract Implementation is IImplementation {
-    address private immutable implementationAddress;
-
     /**
-     * @dev Contract constructor that sets the implementation address to the address of this contract.
+     * @dev Sets the implementation address to the address of this contract.
```
milapsheth, line 14: we decided to revert this file change

contracts/upgradable/Implementation.sol
```diff
@@ -10,14 +10,10 @@ import { IImplementation } from '../interfaces/IImplementation.sol';
  * @dev Derived contracts must implement the setup function.
```
milapsheth, line 10: bump version in package.json

milapsheth (approved):

PR #125, opened 2023-12-19
feat(InterchainMultisig): weighted multisig contract

re1ro:
[AXE-2727](https://axelarnetwork.atlassian.net/browse/AXE-2727)
[AXE-3078](https://axelarnetwork.atlassian.net/browse/AXE-3078)
* [x] `BaseWeightedMultisig`: new abstract contact
* [x] `InterchainMultisig`: chain aware weighted multisig

[AXE-2727]: https://axelarnetwork.atlassian.net/browse/AXE-2727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AXE-3078]: https://axelarnetwork.atlassian.net/browse/AXE-3078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -40,7 +40,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @return bytes32 The signers hash for the given epoch
      */
     function signerHashByEpoch(uint256 epoch) external view returns (bytes32) {
-        return _baseWeightedStorage().hashForSignersEpoch[epoch];
+        return _baseWeightedStorage().signerHashByEpoch[epoch];
     }
 
     /*
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -39,7 +39,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @param epoch The given epoch
      * @return bytes32 The signers hash for the given epoch
      */
-    function hashForSignersEpoch(uint256 epoch) external view returns (bytes32) {
+    function signerHashByEpoch(uint256 epoch) external view returns (bytes32) {
         return _baseWeightedStorage().hashForSignersEpoch[epoch];
     }
 
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -69,9 +69,10 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @notice This function takes messageHash and proof data and reverts if proof is invalid
      * @param messageHash The hash of the message that was signed
      * @param proof The multisig proof data
-     * @dev The proof data should have signers, weights, threshold and signatures encoded
-     * @dev The signers and signatures should be sorted by signer address in ascending order
      * @return isLatestSigners True if the proof is from the latest signer set
+     * @dev The proof data should have signers, weights, threshold and signatures encoded
+     *      The signers and signatures should be sorted by signer address in ascending order
+     *      Example: abi.encode([0x11..., 0x22..., 0x33...], [1, 1, 1], 2, [signature1, signature3])
      */
     function validateProof(bytes32 messageHash, bytes calldata proof) public view returns (bool isLatestSigners) {
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -59,23 +59,27 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @param signatures The signatures data
      * @return isLatestSigners True if provided signers are the current ones
      */
-    function validateProof(
-        bytes32 messageHash,
-        bytes calldata weightedSigners,
-        bytes[] calldata signatures
-    ) public view returns (bool isLatestSigners) {
+    function validateProof(bytes32 messageHash, bytes calldata proof) public view returns (bool isLatestSigners) {
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
-        bytes32 signersHash = keccak256(weightedSigners);
+
+        if (proof.length < 32) revert InvalidSigners();
+
+        WeightedSigners memory signers;
+        bytes[] memory signatures;
+
+        (signers.accounts, signers.weights, signers.threshold, signatures) = abi.decode(
+            proof,
+            (address[], uint256[], uint256, bytes[])
+        );
+
+        bytes32 signersHash = keccak256(abi.encode(signers.accounts, signers.weights, signers.threshold));
         uint256 epoch = slot.signersEpochForHash[signersHash];
         uint256 currentEpoch = slot.currentSignersEpoch;
 
         isLatestSigners = epoch == currentEpoch;
 
         if (signatures.length == 0) revert MalformedSignatures();
-        if (epoch == 0 || currentEpoch - epoch > OLD_SIGNERS_RETENTION || weightedSigners.length == 0)
-            revert InvalidSigners();
-
-        WeightedSigners memory signers = abi.decode(weightedSigners, (WeightedSigners));
+        if (epoch == 0 || currentEpoch - epoch > OLD_SIGNERS_RETENTION) revert InvalidSigners();
 
         _validateSignatures(messageHash, signers, signatures);
     }
@@ -88,26 +92,29 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @notice This function rotates the current signers with a new set of signers
      * @param newWeightedSigners The new weighted signers data
      */
-    function _rotateSigners(WeightedSigners memory newSet) internal {
+    function _rotateSigners(WeightedSigners memory newSigners) internal {
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
 
-        // signers must be sorted binary or alphabetically in lower case
-        if (newSet.signers.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSet.signers)) revert InvalidSigners();
-
-        uint256 length = newSet.signers.length;
+        uint256 length = newSigners.accounts.length;
         uint256 totalWeight;
 
+        // signers must be sorted binary or alphabetically in lower case
+        if (newSigners.accounts.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSigners.accounts))
+            revert InvalidSigners();
+
+        if (newSigners.weights.length != length) revert InvalidWeights();
+
         for (uint256 i; i < length; ++i) {
-            uint256 weight = newSet.signers[i].weight;
+            uint256 weight = newSigners.weights[i];
 
             if (weight == 0) revert InvalidWeights();
 
             totalWeight = totalWeight + weight;
         }
 
-        if (newSet.threshold == 0 || totalWeight < newSet.threshold) revert InvalidThreshold();
+        if (newSigners.threshold == 0 || totalWeight < newSigners.threshold) revert InvalidThreshold();
 
-        bytes32 newSignersHash = keccak256(abi.encode(newSet));
+        bytes32 newSignersHash = keccak256(abi.encode(newSigners.accounts, newSigners.weights, newSigners.threshold));
 
         uint256 epoch = slot.currentSignersEpoch + 1;
         // slither-disable-next-line costly-loop
@@ -115,7 +122,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         slot.hashForSignersEpoch[epoch] = newSignersHash;
         slot.signersEpochForHash[newSignersHash] = epoch;
 
-        emit SignersRotated(newSet);
+        emit SignersRotated(newSigners);
     }
 
     /**********************\
@@ -133,7 +140,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         WeightedSigners memory weightedSigners,
         bytes[] memory signatures
     ) internal pure {
-        uint256 signersLength = weightedSigners.signers.length;
+        uint256 signersLength = weightedSigners.accounts.length;
         uint256 signaturesLength = signatures.length;
         uint256 signerIndex;
         uint256 totalWeight;
@@ -145,17 +152,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             address signer = ECDSA.recover(messageHash, signatures[i]);
 
             // looping through remaining signers to find a match
-            for (
-                ;
-                signerIndex < signersLength && signer != weightedSigners.signers[signerIndex].account;
-                ++signerIndex
-            ) {}
+            for (; signerIndex < signersLength && signer != weightedSigners.accounts[signerIndex]; ++signerIndex) {}
 
             // checking if we are out of signers
             if (signerIndex == signersLength) revert MalformedSignatures();
 
             // accumulating signatures weight
-            totalWeight = totalWeight + weightedSigners.signers[signerIndex].weight;
+            totalWeight = totalWeight + weightedSigners.weights[signerIndex];
 
             // weight needs to reach or surpass threshold
             if (totalWeight >= weightedSigners.threshold) return;
@@ -172,14 +175,14 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @param signers The signers to check
      * @return True if the signers are sorted and contain no duplicates
      */
-    function _isSortedAscAndContainsNoDuplicate(WeightedSigner[] memory signers) internal pure returns (bool) {
+    function _isSortedAscAndContainsNoDuplicate(address[] memory signers) internal pure returns (bool) {
         uint256 signersLength = signers.length;
-        address prevSigner = signers[0].account;
+        address prevSigner = signers[0];
 
         if (prevSigner == address(0)) return false;
 
         for (uint256 i = 1; i < signersLength; ++i) {
-            address currSigner = signers[i].account;
+            address currSigner = signers[i];
 
             if (prevSigner >= currSigner) {
                 return false;
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -22,11 +22,11 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         mapping(bytes32 => uint256) epochBySignerHash;
     }
 
-    // @dev Previous signers retention. 0 means only the current signers are valid
-    // @return The number of epochs to keep the signers valid for signature verification
+    /// @dev Previous signers retention. 0 means only the current signers are valid
+    /// @return The number of epochs to keep the signers valid for signature verification
     uint256 public immutable previousSignersRetention;
 
-    // @param previousSignersRetentionEpochs The number of epochs to keep previous signers valid for signature verification
+    /// @param previousSignersRetentionEpochs The number of epochs to keep previous signers valid for signature verification
     constructor(uint256 previousSignersRetentionEpochs) {
         previousSignersRetention = previousSignersRetentionEpochs;
     }
@@ -35,7 +35,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
     |* External Functions *|
     \**********************/
 
-    /*
+    /**
      * @notice This function returns the current signers epoch
      * @return uint256 The current signers epoch
      */
@@ -43,7 +43,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         return _baseWeightedStorage().epoch;
     }
 
-    /*
+    /**
      * @notice This function returns the signers hash for a given epoch
      * @param signerEpoch The given epoch
      * @return bytes32 The signers hash for the given epoch
@@ -52,7 +52,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         return _baseWeightedStorage().signerHashByEpoch[signerEpoch];
     }
 
-    /*
+    /**
      * @notice This function returns the epoch for a given signers hash
      * @param signerHash The signers hash
      * @return uint256 The epoch for the given signers hash
@@ -61,15 +61,16 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         return _baseWeightedStorage().epochBySignerHash[signerHash];
     }
 
-    /*
+    /**
      * @notice This function takes messageHash and proof data and reverts if proof is invalid
      * @param messageHash The hash of the message that was signed
-     * @param weightedSigners The weighted signers data
-     * @param signatures The signatures data
-     * @return isLatestSigners True if provided signers are the current ones
+     * @param proof The multisig proof data
+     * @dev The proof data should have signers, weights, threshold and signatures encoded
+     * @dev The signers and signatures should be sorted by signer address in ascending order
+     * @return isLatestSigners True if the signers are the latest
      */
     function validateProof(bytes32 messageHash, bytes calldata proof) public view returns (bool isLatestSigners) {
-        if (proof.length < 32 * 4) revert InvalidProof();
+        if (proof.length < 32 * 7) revert InvalidProof();
 
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
         // slither-disable-next-line uninitialized-local
@@ -97,9 +98,10 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
     |* Integration Functions *|
     \*************************/
 
-    /*
+    /**
      * @notice This function rotates the current signers with a new set of signers
-     * @param newWeightedSigners The new weighted signers data
+     * @param newSigners The new weighted signers data
+     * @dev The signers should be sorted by signer address in ascending order
      */
     function _rotateSigners(WeightedSigners memory newSigners) internal {
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
@@ -138,11 +140,12 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
     |* Internal Functions *|
     \**********************/
 
-    /*
+    /**
      * @notice This function takes messageHash and proof data and reverts if proof is invalid
      * @param messageHash The hash of the message that was signed
-     * @param weighted The weighted signers data
+     * @param weightedSigners The weighted signers data
      * @param signatures The sorted signatures data
+     * @dev The signers and signatures should be sorted by signer address in ascending order
      */
     function _validateSignatures(
         bytes32 messageHash,
@@ -179,10 +182,10 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         revert LowSignaturesWeight();
     }
 
-    /*
+    /**
      * @notice This function checks if the provided signers are sorted and contain no duplicates
      * @param signers The signers to check
-     * @return True if the signers are sorted and contain no duplicates
+     * @return True if the signers are sorted in ascending order and contain no duplicates
      */
     function _isSortedAscAndContainsNoDuplicate(address[] memory signers) internal pure returns (bool) {
         uint256 signersLength = signers.length;
@@ -203,9 +206,9 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         return true;
     }
 
-    /*
-     * @notice Gets the storage slot for the WeightedMultisigStorage struct
-     * @return the storage slot
+    /**
+     * @notice Gets the specific storage location for preventing upgrade collisions
+     * @return slot containing the WeightedMultisigStorage struct
      */
     function _baseWeightedStorage() private pure returns (WeightedMultisigStorage storage slot) {
         assembly {
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -26,7 +26,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @notice This function returns the current signers epoch
      * @return uint256 The current signers epoch
      */
-    function currentSignersEpoch() external view returns (uint256) {
+    function signersEpoch() external view returns (uint256) {
         return _baseWeightedStorage().currentSignersEpoch;
     }
 
@@ -48,9 +48,6 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         return _baseWeightedStorage().signersEpochForHash[hash];
     }
 
-    /*************************\
-    |* Integration Functions *|
-    \*************************/
     /*
      * @notice This function takes messageHash and proof data and reverts if proof is invalid
      * @param messageHash The hash of the message that was signed
@@ -58,33 +55,37 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @param signatures The signatures data
      * @return isLatestSigners True if provided signers are the current ones
      */
-    function _validateProof(
+    function validateProof(
         bytes32 messageHash,
         bytes calldata weightedSigners,
         bytes[] calldata signatures
-    ) internal view returns (bool isLatestSigners) {
-        WeightedMultisigStorage storage $ = _baseWeightedStorage();
+    ) public view returns (bool isLatestSigners) {
+        WeightedMultisigStorage storage slot = _baseWeightedStorage();
         bytes32 signersHash = keccak256(weightedSigners);
-        uint256 signersEpoch = $.signersEpochForHash[signersHash];
-        uint256 currentEpoch = $.currentSignersEpoch;
+        uint256 epoch = slot.signersEpochForHash[signersHash];
+        uint256 currentEpoch = slot.currentSignersEpoch;
 
-        isLatestSigners = signersEpoch == currentEpoch;
+        isLatestSigners = epoch == currentEpoch;
 
         if (signatures.length == 0) revert MalformedSignatures();
-        if (signersEpoch == 0 || currentEpoch - signersEpoch >= OLD_SIGNERS_RETENTION || weightedSigners.length == 0) revert InvalidSigners();
+        if (epoch == 0 || currentEpoch - epoch >= OLD_SIGNERS_RETENTION || weightedSigners.length == 0)
+            revert InvalidSigners();
 
         WeightedSigners memory signers = abi.decode(weightedSigners, (WeightedSigners));
 
         _validateSignatures(messageHash, signers, signatures);
     }
 
+    /*************************\
+    |* Integration Functions *|
+    \*************************/
+
     /*
      * @notice This function rotates the current signers with a new set of signers
      * @param newWeightedSigners The new weighted signers data
      */
-    function _rotateSigners(bytes memory newWeightedSigners) internal {
-        WeightedMultisigStorage storage $ = _baseWeightedStorage();
-        WeightedSigners memory newSet = abi.decode(newWeightedSigners, (WeightedSigners));
+    function _rotateSigners(WeightedSigners memory newSet) internal {
+        WeightedMultisigStorage storage slot = _baseWeightedStorage();
 
         // signers must be sorted binary or alphabetically in lower case
         if (newSet.signers.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSet.signers)) revert InvalidSigners();
@@ -102,13 +103,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
 
         if (newSet.threshold == 0 || totalWeight < newSet.threshold) revert InvalidThreshold();
 
-        bytes32 newSignersHash = keccak256(newWeightedSigners);
+        bytes32 newSignersHash = keccak256(abi.encode(newSet));
 
-        uint256 epoch = $.currentSignersEpoch + 1;
+        uint256 epoch = slot.currentSignersEpoch + 1;
         // slither-disable-next-line costly-loop
-        $.currentSignersEpoch = epoch;
-        $.hashForSignersEpoch[epoch] = newSignersHash;
-        $.signersEpochForHash[newSignersHash] = epoch;
+        slot.currentSignersEpoch = epoch;
+        slot.hashForSignersEpoch[epoch] = newSignersHash;
+        slot.signersEpochForHash[newSignersHash] = epoch;
 
         emit SignersRotated(newSet);
     }
@@ -157,19 +158,19 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @return True if the signers are sorted and contain no duplicates
      */
     function _isSortedAscAndContainsNoDuplicate(WeightedSigner[] memory signers) internal pure returns (bool) {
-        uint256 accountsLength = signers.length;
-        address prevAccount = signers[0].account;
+        uint256 signersLength = signers.length;
+        address prevSigner = signers[0].account;
 
-        if (prevAccount == address(0)) return false;
+        if (prevSigner == address(0)) return false;
 
-        for (uint256 i = 1; i < accountsLength; ++i) {
-            address currAccount = signers[i].account;
+        for (uint256 i = 1; i < signersLength; ++i) {
+            address currSigner = signers[i].account;
 
-            if (prevAccount >= currAccount) {
+            if (prevSigner >= currSigner) {
                 return false;
             }
 
-            prevAccount = currAccount;
+            prevSigner = currSigner;
         }
 
         return true;
@@ -179,9 +180,9 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @notice Gets the storage slot for the WeightedMultisigStorage struct
      * @return the storage slot
      */
-    function _baseWeightedStorage() private pure returns (WeightedMultisigStorage storage $) {
+    function _baseWeightedStorage() private pure returns (WeightedMultisigStorage storage slot) {
         assembly {
-            $.slot := BASE_WEIGHTED_STORAGE_LOCATION
+            slot.slot := BASE_WEIGHTED_STORAGE_LOCATION
         }
     }
 }
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -145,7 +145,11 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             address signer = ECDSA.recover(messageHash, signatures[i]);
 
             // looping through remaining signers to find a match
-            for (; signerIndex < signersLength && signer != weightedSigners.signers[signerIndex].account; ++signerIndex) {}
+            for (
+                ;
+                signerIndex < signersLength && signer != weightedSigners.signers[signerIndex].account;
+                ++signerIndex
+            ) {}
 
             // checking if we are out of signers
             if (signerIndex == signersLength) revert MalformedSignatures();
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -99,7 +99,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
         uint256 totalWeight;
 
         // signers must be sorted binary or alphabetically in lower case
-        if (newSigners.accounts.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSigners.accounts))
+        if (length == 0 || !_isSortedAscAndContainsNoDuplicate(newSigners.accounts))
             revert InvalidSigners();
 
         if (newSigners.weights.length != length) revert InvalidWeights();
--- a/contracts/governance/BaseWeightedMultisig.sol
+++ b/contracts/governance/BaseWeightedMultisig.sol
@@ -5,6 +5,10 @@ pragma solidity ^0.8.0;
 import { IBaseWeightedMultisig } from '../interfaces/IBaseWeightedMultisig.sol';
 import { ECDSA } from '../libs/ECDSA.sol';
 
+/**
+    @title BaseWeightedMultisig Contract
+    @notice Base contract to build a weighted multisig verification
+*/
 abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
     // keccak256('WeightedMultisig.Storage');
     bytes32 internal constant BASE_WEIGHTED_STORAGE_LOCATION =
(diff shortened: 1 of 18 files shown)


contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * @notice Withdraws native token from the contract.
+     * @notice This function is protected by the onlySigners modifier.
+     * @param transferData The data for the transfer
+     * @param weightedSigners The weighted signers data
+     * @param signatures The signatures data
+     * @dev This function is only callable by the contract itself after passing according proposal
+     */
+    function withdraw(
```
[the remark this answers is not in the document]
re1ro: Very good point

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,184 @@
...
+        uint256 length = newSet.signers.length;
+        uint256 totalWeight;
+        for (uint256 i; i < length; ++i) {
+            uint256 weight = newSet.signers[i].weight;
+
+            if (weight == 0) revert InvalidWeights();
+
+            totalWeight = totalWeight + weight;
+        }
+        if (newSet.threshold == 0 || totalWeight < newSet.threshold) revert InvalidThreshold();
```
deanamiel, line 100: This condition `newSet.threshold == 0` could be checked earlier to avoid calling the internal function and performing the loop if the function will revert anyways

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,184 @@
...
+     * @param weightedSigners The weighted signers data
+     * @param signatures The signatures data
+     * @return isLatestSigners True if provided signers are the current ones
+     */
+    function _validateProof(
+        bytes32 messageHash,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) internal view returns (bool isLatestSigners) {
+        WeightedMultisigStorage storage $ = _baseWeightedStorage();
```
deanamiel, line 67: Any reason why you named the variable `$`? I've never seen that convention before

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,184 @@
...
+
+    /**********************\
+    |* External Functions *|
+    \**********************/
+
+    /*
+     * @dev This function checks if the provided signers are sorted and contain no duplicates
+     * @param signers The signers to check
+     * @return True if the signers are sorted and contain no duplicates
+     */
```
deanamiel, line 29:
```suggestion
    /*
     * @dev This function returns the current signers epoch
     * @return uint256 The current signers epoch
     */
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,184 @@
...
+    function currentSignersEpoch() external view returns (uint256) {
+        return _baseWeightedStorage().currentSignersEpoch;
+    }
+
+    /*
+     * @dev This function checks if the provided signers are sorted and contain no duplicates
+     * @param signers The signers to check
+     * @return True if the signers are sorted and contain no duplicates
+     */
```
deanamiel, line 38:
```suggestion
    /*
     * @dev This function returns the signers hash for a given epoch
     * @param epoch The given epoch
     * @return bytes32 The signers hash for the given epoch
     */
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,184 @@
...
+    function hashForSignersEpoch(uint256 epoch) external view returns (bytes32) {
+        return _baseWeightedStorage().hashForSignersEpoch[epoch];
+    }
+
+    /*
+     * @dev This function checks if the provided signers are sorted and contain no duplicates
+     * @param signers The signers to check
+     * @return True if the signers are sorted and contain no duplicates
+     */
```
deanamiel, line 47:
```suggestion
    /*
     * @dev This function returns the epoch for a given signers hash
     * @param hash The signers hash
     * @return uint256 The epoch for the given signers hash
     */
```

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,174 @@
...
+     * @dev Calls a target address with specified calldata and passing provided native value.
+     * @param batch The batch of calls to execute
+     * @param weightedSigners The weighted signers payload
+     * @param signatures The signatures payload
+     */
+    function executeCalls(
+        bytes calldata batch,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) external payable {
```
deanamiel, line 66: Add sanity check to make sure signatures is not empty before proceeding?

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,174 @@
...
+        if (target != address(this)) revert InvalidTarget();
+        if ($.isPayloadExecuted[payloadHash]) revert AlreadyExecuted();
+        $.isPayloadExecuted[payloadHash] = true;
+
+        _rotateSigners(newWeightedSigners);
+    }
+
+    /**
+     * @notice Withdraws native token from the contract.
+     * @notice This function is protected by the onlySigners modifier.
```
deanamiel, line 127:
```suggestion
     * @notice This function is protected by the _validateProof internal function.
```

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,174 @@
...
+     * @param payloadHash The hash of the payload payload
+     * @return True if the payload has been executed
+     */
+    function isPayloadExecuted(bytes32 payloadHash) external view returns (bool) {
+        return _interchainMultisigStorage().isPayloadExecuted[payloadHash];
+    }
+
+    /**
+     * @notice Executes an external contract call.
+     * @notice This function is protected by the onlySigners requirement.
```
deanamiel, line 56:
```suggestion
     * @notice This function is protected by the _validateProof internal function.
```

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,174 @@
...
+     * @param signatures The signatures payload
+     * @dev This function is only callable by the contract itself after passing according proposal
+     */
+    function withdraw(
+        bytes calldata transferPayload,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) external payable {
+        bytes32 payloadHash = keccak256(transferPayload);
+        _validateProof(keccak256(transferPayload), weightedSigners, signatures);
```
deanamiel, line 139:
```suggestion
        _validateProof(payloadHash, weightedSigners, signatures);
```

contracts/governance/BaseWeightedMultisig.sol
re1ro: I totally agree but I think we would prioritize code readability over some gas optimization for a revert scenario. Revert is usually caught by the gas estimation for a TX, so normally no gas is wasted

contracts/governance/BaseWeightedMultisig.sol
re1ro: I was trying to follow the convention suggested in the standard:
https://eips.ethereum.org/EIPS/eip-7201

contracts/governance/InterchainMultisig.sol
re1ro: Added in base contract

contracts/interfaces/IInterchainMultisig.sol
```diff
@@ -0,0 +1,49 @@
...
+    error NotSelf();
+    error InvalidProof();
+    error AlreadyExecuted();
+    error InvalidPayloadType();
+    error InvalidChainNameHash();
+    error InvalidTarget();
+    error InvalidRecipient();
+
+    struct InterCall {
+        bytes32 chainNameHash;
```
milapsheth, line 22: string chain would be nicer for debugging. gas optimization is less important for this contract

contracts/interfaces/IInterchainMultisig.sol
```diff
@@ -0,0 +1,49 @@
...
+interface IInterchainMultisig is ICaller, IBaseWeightedMultisig {
+    error NotSelf();
+    error InvalidProof();
+    error AlreadyExecuted();
+    error InvalidPayloadType();
+    error InvalidChainNameHash();
+    error InvalidTarget();
+    error InvalidRecipient();
+
+    struct InterCall {
```
milapsheth, line 21:
```suggestion
    struct Call {
```
Or InterchainCall, although it's not really being sent interchain (so I prefer the above)

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+                bytes4 selector = abi.decode(call.callData, (bytes4));
+                // check if the call is for signers rotation
+                if (call.target == address(this) && selector == InterchainMultisig.rotateSigners.selector)
+                    if (isLatestSigners)
+                        // check if the call is from the latest signers and if so, mark them as not latest
+                        isLatestSigners = false;
+                        // or skip the call
+                    else continue;
+
+                _call(call.target, call.callData, call.nativeValue);
```
milapsheth, line 94: more newlines between lines

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+            // check if the call is for this contract and chain
+            if (call.chainNameHash == chainNameHash && call.caller == address(this)) {
+                bytes4 selector = abi.decode(call.callData, (bytes4));
+                // check if the call is for signers rotation
+                if (call.target == address(this) && selector == InterchainMultisig.rotateSigners.selector)
+                    if (isLatestSigners)
+                        // check if the call is from the latest signers and if so, mark them as not latest
+                        isLatestSigners = false;
+                        // or skip the call
+                    else continue;
```
milapsheth, line 92: put if blocks in {} for readability and less bug risk

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+    ) external payable {
+        bytes32 payloadHash = keccak256(batch);
+        bool isLatestSigners = _validateProof(payloadHash, weightedSigners, signatures);
+
+        InterchainMultisigStorage storage $ = _interchainMultisigStorage();
+        (
+            ,
+            /* bytes32 salt */
+            InterCall[] memory calls
+        ) = abi.decode(batch, (bytes32, InterCall[]));
```
milapsheth, line 75: make nonce into another argument of the function instead, so second arg can be array of Calls? for signature verification use the encoded value

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+        uint256 length = calls.length;
+
+        if ($.isPayloadExecuted[payloadHash]) revert AlreadyExecuted();
+        $.isPayloadExecuted[payloadHash] = true;
```
milapsheth, line 79: Check if nonce is executed instead?

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @param batch The batch of calls to execute
+     * @param weightedSigners The weighted signers payload
+     * @param signatures The signatures payload
+     */
+    function executeCalls(
+        bytes calldata batch,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) external payable {
+        bytes32 payloadHash = keccak256(batch);
```
milapsheth, line 67: use toEthSignedMessage conversion like the gateway?

milapsheth, line 67:
```suggestion
        bytes32 messageHash = ECDSA.toEthSignedMessageHash(keccak256(batch));
```

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+        (
+            ,
+            /* bytes32 salt */
+            InterCall[] memory calls
+        ) = abi.decode(batch, (bytes32, InterCall[]));
+        uint256 length = calls.length;
+
+        if ($.isPayloadExecuted[payloadHash]) revert AlreadyExecuted();
+        $.isPayloadExecuted[payloadHash] = true;
+
```
milapsheth, line 80: emit Executed event for the nonce and batch hash

contracts/governance/InterchainMultisig.sol
milapsheth, line 94: emit event for each Call? useful for monitoring

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @param weightedSigners The weighted signers payload
+     * @param signatures The signatures payload
+     */
+    function executeCalls(
+        bytes calldata batch,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) external payable {
+        bytes32 payloadHash = keccak256(batch);
+        bool isLatestSigners = _validateProof(payloadHash, weightedSigners, signatures);
```
milapsheth, line 68: since we will set the retention to 1 (always pass 1 in the constructor to base contract), and only allow latest signers, the complicated rotation handling logic isn't needed. You can add an unit test to verify old signer can't execute.

contracts/interfaces/IInterchainMultisig.sol
```diff
@@ -0,0 +1,49 @@
...
+        bytes32 chainNameHash;
+        address caller;
```
milapsheth, line 23:
```suggestion
        address executor;
```
or `multisig` perhaps

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+        uint256 length = calls.length;
+
+        if ($.isPayloadExecuted[payloadHash]) revert AlreadyExecuted();
+        $.isPayloadExecuted[payloadHash] = true;
+
+        for (uint256 i; i < length; ++i) {
+            InterCall memory call = calls[i];
+            // check if the call is for this contract and chain
+            if (call.chainNameHash == chainNameHash && call.caller == address(this)) {
+                bytes4 selector = abi.decode(call.callData, (bytes4));
```
milapsheth, line 85: can be simplified with retention set to 1

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        bytes32 newSignersHash = keccak256(newWeightedSigners);
+
+        uint256 epoch = $.currentSignersEpoch + 1;
+        // slither-disable-next-line costly-loop
+        $.currentSignersEpoch = epoch;
+        $.hashForSignersEpoch[epoch] = newSignersHash;
+        $.signersEpochForHash[newSignersHash] = epoch;
+
+        emit SignersRotated(newSet);
```
milapsheth, line 111: emit earlier and include the old set as well?

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+        revert LowSignaturesWeight();
+    }
+
+    /*
+     * @notice This function checks if the provided signers are sorted and contain no duplicates
+     * @param signers The signers to check
+     * @return True if the signers are sorted and contain no duplicates
+     */
+    function _isSortedAscAndContainsNoDuplicate(WeightedSigner[] memory signers) internal pure returns (bool) {
+        uint256 accountsLength = signers.length;
```
milapsheth, line 158: account -> signer everywhere

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+        WeightedSigners memory signers = abi.decode(weightedSigners, (WeightedSigners));
+
+        _validateSignatures(messageHash, signers, signatures);
+    }
+
+    /*
+     * @notice This function rotates the current signers with a new set of signers
+     * @param newWeightedSigners The new weighted signers data
+     */
+    function _rotateSigners(bytes memory newWeightedSigners) internal {
```
milapsheth, line 85: pass struct directly?

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+        uint256 length = newSet.signers.length;
+        uint256 totalWeight;
+        for (uint256 i; i < length; ++i) {
+            uint256 weight = newSet.signers[i].weight;
+
+            if (weight == 0) revert InvalidWeights();
+
+            totalWeight = totalWeight + weight;
+        }
+        if (newSet.threshold == 0 || totalWeight < newSet.threshold) revert InvalidThreshold();
```
milapsheth, line 101:
```suggestion

        if (newSet.threshold == 0 || totalWeight < newSet.threshold) revert InvalidThreshold();
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+    function _rotateSigners(bytes memory newWeightedSigners) internal {
+        WeightedMultisigStorage storage $ = _baseWeightedStorage();
+        WeightedSigners memory newSet = abi.decode(newWeightedSigners, (WeightedSigners));
+
+        // signers must be sorted binary or alphabetically in lower case
+        if (newSet.signers.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSet.signers)) revert InvalidSigners();
+
+        uint256 length = newSet.signers.length;
+        uint256 totalWeight;
+        for (uint256 i; i < length; ++i) {
```
milapsheth, line 94:
```suggestion

        for (uint256 i; i < length; ++i) {
```

contracts/governance/BaseWeightedMultisig.sol
re1ro: I agree that looks like jQuery 🤣  Yeah `storage` sounds good.

contracts/governance/BaseWeightedMultisig.sol
re1ro: There is no old set, as we only have a commit for it. And I think we don't need to emit earlier as there is no function calls made prior to the emit.

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    /**********************\
+    |* External Functions *|
+    \**********************/
+
+    /*
+     * @notice This function returns the current signers epoch
+     * @return uint256 The current signers epoch
+     */
+    function currentSignersEpoch() external view returns (uint256) {
```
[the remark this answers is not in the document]
re1ro: As a base contract that might too generic of a name and could conflict with some other logic build on top of it. How about `signersEpoch` ?

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IBaseWeightedMultisig } from '../interfaces/IBaseWeightedMultisig.sol';
+import { ECDSA } from '../libs/ECDSA.sol';
+
+abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
```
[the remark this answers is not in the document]
re1ro: Maybe we should have this as a base contract for the GMP Gateway instead of external module. We had code size issues with the original gateway and had to make it a module. 

In order for it to be a module it also needs to be an Ownable and have a little different setup

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,185 @@
...
+    // keccak256('WeightedMultisig.Storage');
+    bytes32 private constant BASE_WEIGHTED_STORAGE_LOCATION =
```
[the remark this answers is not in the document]
re1ro: It's not optimized for the assembly usage so we have to keep it hardcoded

contracts/governance/BaseWeightedMultisig.sol
re1ro: Yeah that makes the code cleaner but I now I need to encode them again for hashing. I think it's not too much of an overhead.

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @notice Executes an external contract call.
+     * @notice This function is protected by the onlySigners requirement.
+     * @dev Calls a target address with specified calldata and passing provided native value.
+     * @param batch The batch of calls to execute
+     * @param weightedSigners The weighted signers payload
+     * @param signatures The signatures payload
+     */
+    function executeCalls(
+        bytes calldata batch,
+        bytes calldata weightedSigners,
```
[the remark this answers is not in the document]
re1ro: because we use the blob for hashing. Otherwise we will need to re-encode the struct again to hash it

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+import { Caller } from '../utils/Caller.sol';
+
+/**
+ * @title Multisig Contract
+ * @notice An extension of MultisigBase that can call functions on any contract.
+ */
+contract InterchainMultisig is Caller, BaseWeightedMultisig, IInterchainMultisig {
+    // keccak256('InterchainMultisig.Storage')
+    bytes32 private constant INTERCHAIN_MULTISIG_INTERCHAINMULTISIGSTORAGE_LOCATION =
+        0x5a9dc2248a56f285e6221da431581a5990380ebfa07727563571b2be1517a23e;
```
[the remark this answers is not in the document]
re1ro: Doesn't work for assembly usage

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @notice This function is protected by the onlySigners requirement.
+     * @dev Calls a target address with specified calldata and passing provided native value.
+     * @param batch The batch of calls to execute
+     * @param weightedSigners The weighted signers payload
+     * @param signatures The signatures payload
+     */
+    function executeCalls(
+        bytes calldata batch,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
```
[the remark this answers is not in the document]
re1ro: We need `weightedSigners` as bytes for efficiency

contracts/governance/InterchainMultisig.sol
re1ro: Nonce is part of the signed data. I would need to abi.encode all the data again for hashing which is inefficient

contracts/governance/InterchainMultisig.sol
re1ro: Same salt could be used twice for different payload. Otherwise we would need to track used nonces somewhere

Foivos (approved):

deanamiel (approved):
Let me know if you want some help with test coverage

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -59,23 +59,27 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
...
-    function validateProof(
-        bytes32 messageHash,
-        bytes calldata weightedSigners,
-        bytes[] calldata signatures
-    ) public view returns (bool isLatestSigners) {
+    function validateProof(bytes32 messageHash, bytes calldata proof) public view returns (bool isLatestSigners) {
         WeightedMultisigStorage storage slot = _baseWeightedStorage();
-        bytes32 signersHash = keccak256(weightedSigners);
+
+        if (proof.length < 32) revert InvalidSigners();
```
milapsheth, line 65: Is this needed? Can we rely on abi.decode to fail instead?

contracts/governance/BaseWeightedMultisig.sol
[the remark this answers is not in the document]
milapsheth: i see, needs a test then

contracts/interfaces/IBaseWeightedMultisig.sol
```diff
@@ -0,0 +1,54 @@
...
+pragma solidity ^0.8.0;
+
+interface IBaseWeightedMultisig {
+    error InvalidSigners();
+    error InvalidThreshold();
+    error MalformedSignatures();
+    error LowSignaturesWeight();
+    error InvalidWeights();
+
+    struct WeightedSigners {
```
milapsheth, line 12: Since we're not exposing this in the API, this should be defined in the implementation instead

contracts/interfaces/IBaseWeightedMultisig.sol
```diff
@@ -0,0 +1,54 @@
...
+        address[] accounts;
+        uint256[] weights;
+        uint256 threshold;
+    }
+
+    event SignersRotated(WeightedSigners signers);
```
milapsheth, line 18: use individual fields for consistency

contracts/interfaces/IBaseWeightedMultisig.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @dev This function returns the old signers retention period
+     * @return uint256 The old signers retention period
+     */
+    function OLD_SIGNERS_RETENTION() external view returns (uint256);
```
milapsheth, line 24: linter warning

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,126 @@
...
+        bytes32 messageHash = ECDSA.toEthSignedMessageHash(keccak256(callBatch));
+        validateProof(messageHash, proof);
+
+        InterchainMultisigStorage storage slot = _interchainMultisigStorage();
+        (bytes32 salt, Call[] memory calls) = abi.decode(callBatch, (bytes32, Call[]));
+        uint256 length = calls.length;
+
+        emit BatchExecuted(messageHash, salt, length);
+
+        if (slot.isPayloadExecuted[messageHash]) revert AlreadyExecuted();
```
milapsheth, line 72:
```suggestion
        if (slot.isBatchExecuted[messageHash]) revert AlreadyExecuted();
```
or `isMessageExecuted`

We're using 3 different terms to represent the same thing which is confusing. Let's stick to batch?

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @notice Withdraws native token from the contract.
+     * @notice This function is protected by the onlySelf modifier.
+     * @param recipient The recipient of the native value
+     * @param amount The amount of native value to withdraw
+     * @dev This function is only callable by the contract itself after signature verification
+     */
+    function withdraw(address recipient, uint256 amount) external payable onlySelf {
+        if (recipient == address(0)) revert InvalidRecipient();
+        if (amount > address(this).balance) revert InsufficientBalance();
```
milapsheth, line 108:
```suggestion

        if (amount > address(this).balance) revert InsufficientBalance();
```

contracts/interfaces/IBaseWeightedMultisig.sol
```diff
@@ -0,0 +1,54 @@
...
+
+interface IBaseWeightedMultisig {
+    error InvalidSigners();
+    error InvalidThreshold();
+    error MalformedSignatures();
+    error LowSignaturesWeight();
+    error InvalidWeights();
+
+    struct WeightedSigners {
+        address[] accounts;
```
milapsheth, line 13:
```suggestion
        address[] signers;
```
to avoid both terms being used

contracts/interfaces/IInterchainMultisig.sol
```diff
@@ -0,0 +1,48 @@
...
+    event BatchExecuted(bytes32 indexed messageHash, bytes32 indexed salt, uint256 indexed length);
+
+    event CallExecuted(bytes32 indexed messageHash, address indexed target, bytes callData, uint256 nativeValue);
+
+    /**
+     * @notice Checks if a payload has been executed
+     * @param payloadHash The hash of the payload payload
+     * @return True if the payload has been executed
+     */
+    function isPayloadExecuted(bytes32 payloadHash) external view returns (bool);
```
milapsheth, line 38:
```suggestion
    function isBatchExecuted(bytes32 batchHash) external view returns (bool);
```

contracts/interfaces/IInterchainMultisig.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @notice Executes an external contract call.
+     * @notice This function is protected by the onlySigners requirement.
+     * @dev Calls a target address with specified calldata and passing provided native value.
+     * @param batch The batch of calls to execute
+     * @param proof The multisig proof data
+     */
+    function executeCalls(bytes calldata batch, bytes calldata proof) external payable;
```
milapsheth, line 47: Would be preferable to use a Call array here instead. Gas cost isn't super concerning for multisig

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,206 @@
...
+     * @param newWeightedSigners The new weighted signers data
+     */
+    function _rotateSigners(WeightedSigners memory newSigners) internal {
+        WeightedMultisigStorage storage slot = _baseWeightedStorage();
+
+        uint256 length = newSigners.accounts.length;
+        uint256 totalWeight;
+
+        // signers must be sorted binary or alphabetically in lower case
+        if (newSigners.accounts.length == 0 || !_isSortedAscAndContainsNoDuplicate(newSigners.accounts))
```
milapsheth, line 102:
```suggestion
        if (length == 0 || !_isSortedAscAndContainsNoDuplicate(newSigners.accounts))
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,206 @@
...
+
+    /**********************\
+    |* External Functions *|
+    \**********************/
+
+    /*
+     * @notice This function returns the current signers epoch
+     * @return uint256 The current signers epoch
+     */
+    function signersEpoch() external view returns (uint256) {
```
milapsheth, line 33:
```suggestion
    function epoch() external view returns (uint256) {
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,206 @@
...
+    function hashForSignersEpoch(uint256 epoch) external view returns (bytes32) {
+        return _baseWeightedStorage().hashForSignersEpoch[epoch];
+    }
+
+    /*
+     * @notice This function returns the epoch for a given signers hash
+     * @param hash The signers hash
+     * @return uint256 The epoch for the given signers hash
+     */
+    function signersEpochForHash(bytes32 signerHash) external view returns (uint256) {
```
milapsheth, line 51:
```suggestion
    function epochBySignerHash(bytes32 signerHash) external view returns (uint256) {
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,206 @@
...
+
+abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
+    // keccak256('WeightedMultisig.Storage');
+    bytes32 private constant BASE_WEIGHTED_STORAGE_LOCATION =
+        0xa233fbcae4dcfad00091a9d8ff9561f12b3db9ec7227470684b4617d40a38746;
+
+    struct WeightedMultisigStorage {
+        uint256 currentSignersEpoch;
+        mapping(uint256 => bytes32) hashForSignersEpoch;
+        mapping(bytes32 => uint256) signersEpochForHash;
```
milapsheth, line 16:
```suggestion
    struct WeightedMultisigStorage {
        uint256 epoch;
        mapping(uint256 => bytes32) signerHashByEpoch;
        mapping(bytes32 => uint256) epochBySignerHash;
```
`for` usage is less readable

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,206 @@
...
+
+        if (newSigners.threshold == 0 || totalWeight < newSigners.threshold) revert InvalidThreshold();
+
+        bytes32 newSignersHash = keccak256(abi.encode(newSigners.accounts, newSigners.weights, newSigners.threshold));
+
+        uint256 epoch = slot.currentSignersEpoch + 1;
+        // slither-disable-next-line costly-loop
+        slot.currentSignersEpoch = epoch;
+        slot.hashForSignersEpoch[epoch] = newSignersHash;
+        slot.signersEpochForHash[newSignersHash] = epoch;
```
milapsheth, line 123: Add a comment indicating that we overwrite the latest epoch for a signerHash if it's repeated

contracts/governance/BaseWeightedMultisig.sol
re1ro: In this case we will get an ambiguous revert message from EVM. I am ok with removing this check if you want. Or we can actually make the number to be 32 * 4.

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,134 @@
...
+        bytes calldata callBatch,
+        bytes calldata weightedSigners,
+        bytes[] calldata signatures
+    ) external payable {
+        bytes32 messageHash = ECDSA.toEthSignedMessageHash(keccak256(callBatch));
+        validateProof(messageHash, weightedSigners, signatures);
+
+        InterchainMultisigStorage storage slot = _interchainMultisigStorage();
+        (
+            bytes32 salt,
```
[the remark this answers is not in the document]
re1ro: Lets do nonce for clarity of intention.

contracts/interfaces/IInterchainMultisig.sol
re1ro: We can't encode structs in solidity, only decode them. I get this compiler error:
```
TypeError: Type not supported in packed mode.
```

milapsheth (approved):

contracts/test/governance/TestIntrerchainMultisig.sol
```diff
@@ -0,0 +1,15 @@
...
+pragma solidity ^0.8.0;
+
+import { InterchainMultisig } from '../../governance/InterchainMultisig.sol';
+
+contract TestInterchainMultisig is InterchainMultisig {
+    constructor(string memory chainName, WeightedSigners memory weightedSigners)
+        InterchainMultisig(chainName, weightedSigners)
+    {
+        if (INTERCHAIN_MULTISIG_STORAGE != keccak256('InterchainMultisig.Storage')) {
+            revert('Invalid InterchainMultisig.Storage location');
```
milapsheth, line 12: lint warning here. use an error code instead?

test/governance/InterchainMultisig.js
```diff
@@ -0,0 +1,256 @@
...
+        interchainMultisigFactory = await ethers.getContractFactory('TestInterchainMultisig', owner);
+        targetFactory = await ethers.getContractFactory('Target', owner);
+
+        const targetInterface = new Interface(['function callTarget() external']);
+        calldata = targetInterface.encodeFunctionData('callTarget');
+    });
+
+    beforeEach(async () => {
+        // new multisig and target contracts for each test
+        interchainMultisig = await interchainMultisigFactory.deploy('Ethereum', [
```
milapsheth, line 40: many tests can reuse the same multisig contract, group those under one describe block so you can deploy just once. this speeds up live network tests significantly

test/governance/InterchainMultisig.js
```diff
@@ -0,0 +1,256 @@
...
+            threshold,
+        ]);
+
+        await interchainMultisig.deployTransaction.wait(network.config.confirmations);
+
+        targetContract = await targetFactory.deploy().then((d) => d.deployed());
+    });
+
+    it('should revert on execute with insufficient value sent', async () => {
+        const nativeValue = 100;
```
milapsheth, line 52: could set this earlier to reuse

test/governance/InterchainMultisig.js
```diff
@@ -0,0 +1,256 @@
...
+                2,
+                signers,
+            ),
+        );
+
+        const newBalance = await ethers.provider.getBalance(recipient);
+        expect(newBalance).to.equal(oldBalance.add(nativeValue));
+    });
+
+    it('should rotate signers', async () => {
```
milapsheth, line 231: for sanity, add a test that execute call fails from invalid signers

test/governance/BaseWeightedMultisig.js
```diff
@@ -0,0 +1,362 @@
...
+                await getWeightedSignaturesProof(
+                    data,
+                    signers,
+                    signers.map(() => 1),
+                    threshold,
+                    signers.slice(0, threshold),
+                ),
+            );
+
+            expect(isCurrentSigners).to.be.equal(true);
```
deanamiel, line 61:
```suggestion
            expect(isCurrentSigners).to.be.true;
```

test/governance/BaseWeightedMultisig.js
```diff
@@ -0,0 +1,362 @@
...
+
+            await expectRevert(
+                (gasOptions) => multisig.rotateSigners([newSigners, newSigners.map(() => 1), 2], gasOptions),
+                multisig,
+                'InvalidSigners',
+            );
+        });
+
+        it('should not allow transferring signership to address zero', async () => {
+            const newSigners = [ethers.constants.AddressZero, '0x6D4017D4b1DCd36e6EA88b7900e8eC64A1D1315b'];
```
deanamiel, line 262:
```suggestion
            const newSigners = [AddressZero, '0x6D4017D4b1DCd36e6EA88b7900e8eC64A1D1315b'];
```
Can import AddressZero under constants similar to ethers utils

test/governance/InterchainMultisig.js
```diff
@@ -0,0 +1,333 @@
...
+                [call],
+                getWeightedSignaturesProof(
+                    encodeInterchainCallsBatch(formatBytes32String('5'), [call]),
+                    signers,
+                    signers.map(() => 1),
+                    2,
+                    signers,
+                ),
+            ),
+        ).to.be.revertedWithCustomError(interchainMultisig, 'InsufficientBalance');
```
deanamiel, line 70: Use `expertRevert` instead for live testnet compatibility, applies to other instances of `revertedWithCustomError` as well

deanamiel (approved):

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,153 @@
...
+            Call memory call = calls[i];
+
+            // check if the call is for this contract and chain
+            if (keccak256(bytes(call.chainName)) == chainNameHash && call.executor == address(this)) {
+                if (call.target == address(0)) revert InvalidTarget();
+
+                if (call.target == address(this) && bytes4(call.callData) == InterchainMultisig.voidBatch.selector)
+                    if (length == 1) {
+                        emit BatchVoided(batchId);
+                        return;
```
Foivos, line 91: This returns without calling self.voidBatch()

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,153 @@
...
+
+        recipient.safeNativeTransfer(amount);
+    }
+
+    /**
+     * @notice Voids the batch id from being executed in the future.
+     * @notice This function is protected by the onlySelf modifier.
+     * @dev This function is only callable by the contract itself after signature verification
+     */
+    function voidBatch() external payable onlySelf {}
```
Foivos, line 138: When would this get called?

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,153 @@
...
+
+        for (uint256 i; i < length; ++i) {
+            Call memory call = calls[i];
+
+            // check if the call is for this contract and chain
+            if (keccak256(bytes(call.chainName)) == chainNameHash && call.executor == address(this)) {
+                if (call.target == address(0)) revert InvalidTarget();
+
+                if (call.target == address(this) && bytes4(call.callData) == InterchainMultisig.voidBatch.selector)
+                    if (length == 1) {
```
milapsheth, line 89: this doesn't allow voiding for all chains at once. I suggest removing this and renaming `voidBatch` to `noop`, and implicitly let that call be used for voiding. So the logic here just needs to check if at least 1 call for this chain was executed, and revert otherwise. Our scripts can make use of noop for voiding.

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,218 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IBaseWeightedMultisig } from '../interfaces/IBaseWeightedMultisig.sol';
+import { ECDSA } from '../libs/ECDSA.sol';
+
+abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
+    // keccak256('WeightedMultisig.Storage');
+    bytes32 internal constant BASE_WEIGHTED_STORAGE_LOCATION =
```
milapsheth, line 10:
```suggestion
    // keccak256('BaseWeightedMultisig.Storage');
    bytes32 internal constant BASE_WEIGHTED_MULTISIG_SLOT =
```
and update the keccak256 hash perhaps? Or we can change the contract name

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,218 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import { IBaseWeightedMultisig } from '../interfaces/IBaseWeightedMultisig.sol';
+import { ECDSA } from '../libs/ECDSA.sol';
+
+abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
```
milapsheth, line 8: missing natspec on the contract

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,149 @@
...
+        emit BatchExecuted(batchId, messageHash, length);
+
+        uint256 executedCalls = 0;
+
+        for (uint256 i; i < length; ++i) {
+            Call memory call = calls[i];
+
+            // check if the call is for this contract and chain
+            if (keccak256(bytes(call.chainName)) == chainNameHash && call.executor == address(this)) {
+                if (call.target == address(0)) revert InvalidTarget();
```
milapsheth, line 90: `_call` will already handle this check

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: MIT
+
```
milapsheth, line 2: full unit test coverage is still missing, specifically some of the modifier checks (see branch columns etc.)
Screenshot 2024-02-20 at 6 26 33 PM

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,149 @@
...
+        if (slot.isBatchExecuted[batchId]) revert AlreadyExecuted();
+        slot.isBatchExecuted[batchId] = true;
+
+        emit BatchExecuted(batchId, messageHash, length);
+
+        uint256 executedCalls = 0;
+
+        for (uint256 i; i < length; ++i) {
+            Call memory call = calls[i];
```
milapsheth, line 86:
```suggestion
            Call calldata call = calls[i];
```
Switch to calldata since we aren't modifying any variables, and so we also don't need to change `Caller` contract

contracts/governance/InterchainMultisig.sol
```diff
@@ -0,0 +1,149 @@
...
+        InterchainMultisigStorage storage slot = _interchainMultisigStorage();
+        bytes32 messageHash = ECDSA.toEthSignedMessageHash(keccak256(abi.encode(batchId, calls)));
+        uint256 length = calls.length;
+
+        validateProof(messageHash, proof);
+
+        if (slot.isBatchExecuted[batchId]) revert AlreadyExecuted();
+        slot.isBatchExecuted[batchId] = true;
+
+        emit BatchExecuted(batchId, messageHash, length);
```
milapsheth, line 81: maybe emit this at the end, and include executedCalls count as well?

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -0,0 +1,215 @@
...
+            uint256 weight = newSigners.weights[i];
+
+            if (weight == 0) revert InvalidWeights();
+
+            totalWeight = totalWeight + weight;
+        }
+
+        if (newSigners.threshold == 0 || totalWeight < newSigners.threshold) revert InvalidThreshold();
+
+        bytes32 newSignersHash = keccak256(abi.encode(newSigners.signers, newSigners.weights, newSigners.threshold));
```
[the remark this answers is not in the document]
re1ro: As per new requirements we want to allow it now. So the same signer set could be used again later

contracts/governance/InterchainMultisig.sol
re1ro: Changed to simpler design

contracts/governance/InterchainMultisig.sol
milapsheth: still missing a few branches in coverage

contracts/utils/Caller.sol
```diff
@@ -13,13 +13,14 @@ contract Caller is ICaller {
      */
     function _call(
         address target,
-        bytes calldata callData,
+        bytes memory callData,
```
milapsheth, line 16: this change needs to be reverted

contracts/utils/Caller.sol
```diff
@@ -13,13 +13,14 @@ contract Caller is ICaller {
...
         uint256 nativeValue
     ) internal returns (bytes memory) {
         if (!target.isContract()) revert InvalidContract(target);
 
         if (nativeValue > address(this).balance) revert InsufficientBalance();
 
+        // slither-disable-next-line calls-loop
```
milapsheth, line 23: same

test/governance/InterchainGovernance.js
```diff
@@ -386,9 +386,9 @@ describe('InterchainGovernance', () => {
         const bytecodeHash = keccak256(bytecode);
 
         const expected = {
-            istanbul: '0x2534d1533c9ffce84d3174c1f846a4041d07b56d1e7b5cb7138e06fb42086325',
-            berlin: '0x1084d7de843267ed6f4ad87cbdc541bfb2aa003c67c285d0b4f90b3026370f7e',
-            london: '0x9d89dce5b3087d6f9a1b80cc3e96ae9c204a1ce3c2c4eb5bce7671a20a635f97',
+            istanbul: '0x757373424522ec9ac6592d236ac4b1bbdc1f4eadc26a87f8d8d62f95015da89d',
```
milapsheth, line 389: should revert the change

contracts/utils/Caller.sol
re1ro: The problem is EVM can't convert from `memory` to `calldata`. 
If we keep it as `calldata` this will not work:
```
_call(
  0x1..1,
  abi.encodeWithSelector(IERC20.approve.selector, receiver, amount),
  0
)
```
and we might need to change it in future and affect the bytecode of the dependencies

milapsheth: yeah, i understand, but we can change it later if really needed

milapsheth (approved):

PR #126, opened 2024-01-27
feature: error handling

luislucena16:
I add this error handling proposal, I find it cleaner and tidier to work with and also you leave the interface for the specific use i.e. without the errors there, I made sure it compiled!

### DIFF
--- a/contracts/executable/AxelarExecutable.sol
+++ b/contracts/executable/AxelarExecutable.sol
@@ -4,8 +4,9 @@ pragma solidity ^0.8.0;
 
 import { IAxelarGateway } from '../interfaces/IAxelarGateway.sol';
 import { IAxelarExecutable } from '../interfaces/IAxelarExecutable.sol';
+import { AxelarExecutableErrors } from './AxelarExecutableErrors.sol';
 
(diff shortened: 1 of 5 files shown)


re1ro:
Hi @luislucena16! 
Thank you for the contribution! Usually splitting interfaces done is app level project and not at the library level, so the lib imports are simple. I know it's a preference thing, but are trying to keep it simple as the amount of utils and interfaces keeps growing.

Since the interface itself is tiny and not expected to change you could just have it defined in your project instead of importing.

luislucena16:
hey @re1ro yes, as you said, it is also considered a matter of preference! :) and of course, if the project grows, it is beneficial

luislucena16:
I was thinking about adding more documentation for contributors in the readme, when I was going to do it I didn't see any about it.. what do you think?

milapsheth:
Yeah, we prefer to keep it this way for simplicity

milapsheth:
> I was thinking about adding more documentation for contributors in the readme, when I was going to do it I didn't see any about it.. what do you think?

Sure, additional documentation would be helpful, especially around usage of some our contracts/utils

PR #130, opened 2024-03-01
feat(InterchainMultisig): full interface and script utils

re1ro:
* [x] NPM version bump to `7.0.0`
* [x] `InterchainMultisig`: adding all methods to the interface
* [x] Utils: `getWeightedSignersProof` and `sortWeightedSignaturesProof` 
* [x] `Upgradable`: storing new implementation before the delegate call to satisfy new Slither

### DIFF
--- a/contracts/upgradable/Upgradable.sol
+++ b/contracts/upgradable/Upgradable.sol
@@ -50,11 +50,6 @@ abstract contract Upgradable is Ownable, Implementation, IUpgradable {
             revert InvalidImplementation();
 
         if (newImplementationCodeHash != newImplementation.codehash) revert InvalidCodeHash();
-        if (newImplementation == implementation()) revert InvalidImplementation();
-
-        assembly {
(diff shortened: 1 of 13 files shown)


Foivos (approved):

contracts/upgradable/Upgradable.sol
```diff
@@ -50,6 +50,11 @@ abstract contract Upgradable is Ownable, Implementation, IUpgradable {
             revert InvalidImplementation();
 
         if (newImplementationCodeHash != newImplementation.codehash) revert InvalidCodeHash();
+        if (newImplementation == implementation()) revert InvalidImplementation();
```
milapsheth, line 53: Why do we need this restriction? It can be useful to upgrade to the same implementation for the purpose of calling setup (say to migrate something, such as operators etc.)

milapsheth (approved):

deanamiel (approved):

PR #131, opened 2024-03-05
fix(Upgradable): storing new implementation before the delegate call

re1ro:
* [x] `Upgradable`: storing new implementation before the delegate call to silent Slither warning

### DIFF
--- a/test/upgradable/Upgradable.js
+++ b/test/upgradable/Upgradable.js
@@ -43,20 +43,6 @@ describe('Upgradable', () => {
             expect(await implementation.implementation()).to.eq(AddressZero);
         });
 
-        it('should revert on upgrade with invalid contract ID', async () => {
-            const invalidUpgradableTestFactory = await ethers.getContractFactory('InvalidUpgradableTest', ownerWallet);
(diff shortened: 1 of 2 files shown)


test/upgradable/Upgradable.js
```diff
@@ -26,9 +26,8 @@ describe('Upgradable', () => {
         upgradableTestFactory = await ethers.getContractFactory('UpgradableTest', ownerWallet);
 
         create3DeployerFactory = await ethers.getContractFactory('Create3Deployer', ownerWallet);
-    });
 
-    beforeEach(async () => {
```
milapsheth, line 31: let's avoid reusing the contract instance for positive tests unless the tests are really independent (unlike in this case). you can group revert tests and have a single before block for it. for positive tests, can still have a before each

milapsheth (approved):

test/upgradable/Upgradable.js
re1ro: Sounds good!

deanamiel (approved):

PR #138, opened 2024-03-20
refactor(amplifier): avoid self external calls by reverting on error

milapsheth:
[AXE-3489]

[AXE-3489]: https://axelarnetwork.atlassian.net/browse/AXE-3489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -128,40 +112,43 @@ contract AxelarAmplifierGateway is Ownable, IAxelarAmplifierGateway {
...
-    /******************\
-    |* Self Functions *|
-    \******************/
+    /**********************\
+    |* Internal Functions *|
+    \**********************/
 
-    function approveContractCall(bytes calldata params, bytes32 commandId) external onlySelf {
+    function _approveContractCall(
+        AxelarAmplifierGatewayStorage storage slot,
```
re1ro, line 148: Maybe there is no need to pass the storage slot as it can be accessed in the function by `_axelarAmplifierGatewayStorage()`

re1ro (approved):

Foivos (approved):

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: confirmed it's cheaper to retrieve it directly instead of creating a variable

PR #142, opened 2024-03-21
ci: switch to blacksmith runners

milapsheth:


.github/workflows/codecov.yaml
```diff
@@ -13,22 +13,16 @@ concurrency:
 
 jobs:
   coverage:
-    name: Code Coverage
```
re1ro, line 16: Keep the name?

.github/workflows/lint.yaml
```diff
@@ -9,20 +9,16 @@ concurrency:
 
 jobs:
   lint:
-    strategy:
```
re1ro, line 12: Add name?

.github/workflows/slither.yaml
```diff
@@ -8,21 +8,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  analyze:
```
re1ro, line 11: Name?

.github/workflows/test.yaml
```diff
@@ -9,20 +9,16 @@ concurrency:
 
 jobs:
   test:
-    strategy:
```
re1ro, line 12: Name?

.github/workflows/codecov.yaml
milapsheth: it already shows the name of the workflow from the top of the file

re1ro (approved):

PR #148, opened 2024-03-27
fix(amplifier)!: add messageId and additional domain separation to amplifier gateway

milapsheth:
[AXE-3484]

[AXE-3484]: https://axelarnetwork.atlassian.net/browse/AXE-3484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -14,29 +14,21 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         0xca458dc12368669a3b8c292bc21c1b887ab1aa386fa3fcc1ed972afd74a330ca;
 
     struct AxelarAmplifierGatewayStorage {
-        string chainName;
-        string router;
         mapping(bytes32 => bool) commands;
         mapping(bytes32 => bool) approvals;
     }
 
     IAxelarGatewayWeightedAuth public immutable authModule;
-    bytes32 public immutable chainNameHash;
-    bytes32 public immutable routerHash;
+    bytes32 public immutable domainSeparator;
 
     constructor(
         address authModule_,
-        string memory chainName,
-        string memory router
+        bytes32 domainSeparator_
     ) {
         if (authModule_.code.length == 0) revert InvalidAuthModule();
 
         authModule = IAxelarGatewayWeightedAuth(authModule_);
-
-        _storage().chainName = chainName;
-        _storage().router = router;
-        chainNameHash = keccak256(bytes(chainName));
-        routerHash = keccak256(bytes(router));
+        domainSeparator = domainSeparator_;
     }
 
     /******************\
@@ -114,8 +106,7 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         // returns true for current operators
         bool allowOperatorshipTransfer = authModule.validateProof(batchHash, signedBatch.proof);
 
-        if (chainNameHash != signedBatch.batch.chainNameHash) revert InvalidChainName();
-        if (routerHash != signedBatch.batch.routerHash) revert InvalidRouter();
+        if (domainSeparator != signedBatch.batch.domainSeparator) revert InvalidDomainSeparator();
 
         Command[] memory commands = batch.commands;
 
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -108,10 +108,10 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
 
         if (domainSeparator != signedBatch.batch.domainSeparator) revert InvalidDomainSeparator();
 
-        Command[] memory commands = batch.commands;
+        Command[] calldata commands = batch.commands;
 
         for (uint256 i; i < commands.length; ++i) {
-            Command memory command = commands[i];
+            Command calldata command = commands[i];
             bytes32 commandId = keccak256(bytes(command.messageId));
 
             // Ignore if commandId is already executed
@@ -177,31 +177,32 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         }
     }
 
-    function _approveContractCall(bytes32 commandId, Command memory command) internal {
-        (string memory sourceChain, string memory sourceAddress, address contractAddress, bytes32 payloadHash) = abi
-            .decode(command.params, (string, string, address, bytes32));
+    function _approveContractCall(bytes32 commandId, Command calldata command) internal {
+        ContractCallApprovalParams memory params = abi.decode(command.params, (ContractCallApprovalParams));
 
         bytes32 key = _getIsContractCallApprovedKey(
             commandId,
-            sourceChain,
-            sourceAddress,
-            contractAddress,
-            payloadHash
+            params.sourceChain,
+            params.sourceAddress,
+            params.contractAddress,
+            params.payloadHash
         );
         _storage().approvals[key] = true;
 
         emit ContractCallApproved(
             commandId,
-            sourceChain,
-            sourceAddress,
-            contractAddress,
-            payloadHash,
+            params.sourceChain,
+            params.sourceAddress,
+            params.contractAddress,
+            params.payloadHash,
             command.messageId
         );
     }
 
-    function _transferOperatorship(bytes memory newOperatorsData) internal {
-        authModule.transferOperatorship(newOperatorsData);
+    function _transferOperatorship(bytes calldata newOperatorsData) internal {
+        TransferOperatorshipParams memory params = abi.decode(newOperatorsData, (TransferOperatorshipParams));
+
+        authModule.transferOperatorship(params.newOperators);
 
         // slither-disable-next-line reentrancy-events
         emit OperatorshipTransferred(newOperatorsData);
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -100,8 +100,7 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         // This should be customized per chain.
         bytes32 batchHash = ECDSA.toEthSignedMessageHash(keccak256(data));
 
-        // returns true for current operators
-        bool allowOperatorshipTransfer = authModule.validateProof(batchHash, signedBatch.proof);
+        bool isLatestSigners = authModule.validateProof(batchHash, signedBatch.proof);
 
         if (domainSeparator != signedBatch.batch.domainSeparator) revert InvalidDomainSeparator();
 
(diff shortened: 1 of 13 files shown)


contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -9,6 +9,32 @@ import { IAxelarGMPGateway } from './IAxelarGMPGateway.sol';
...
+
+    struct CommandBatch {
+        bytes32 chainNameHash;
+        bytes32 routerHash;
+        Command[] commands;
+    }
+
+    struct SignedCommandBatch {
+        CommandBatch batch;
+        bytes proof;
```
milapsheth, line 35: Considering if we should type this as well

contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -28,7 +90,41 @@ interface IAxelarAmplifierGateway is IAxelarGMPGateway {
...
+     * @notice Validates and approves a contract call.
+     * @dev Validates the given contract call information and marks it as approved if valid.
+     * @param messageId The unique identifier of the message.
+     * @param sourceChain The name of the source chain.
+     * @param sourceAddress The address of the sender on the source chain.
+     * @param payloadHash The keccak256 hash of the payload data.
+     * @return True if the contract call is validated and approved, false otherwise.
      */
-    function execute(bytes calldata batch) external;
+    function validateMessage(
```
milapsheth, line 124: Maybe `validateContractCallByMessageId` instead

contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -9,6 +9,32 @@ import { IAxelarGMPGateway } from './IAxelarGMPGateway.sol';
...
+
+    struct Command {
+        CommandType commandType;
+        string messageId;
+        bytes params;
+    }
+
+    struct CommandBatch {
+        bytes32 chainNameHash;
+        bytes32 routerHash;
```
milapsheth, line 29: add operator hash commitment that's used to sign the message

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -8,24 +8,35 @@ import { IAxelarGatewayWeightedAuth } from '../interfaces/IAxelarGatewayWeighted
...
+    constructor(
+        address authModule_,
+        string memory chainName,
+        string memory router
+    ) {
         if (authModule_.code.length == 0) revert InvalidAuthModule();
 
         authModule = IAxelarGatewayWeightedAuth(authModule_);
+
+        _storage().chainName = chainName;
```
re1ro, line 36: Maybe cache the slot? And call it gatewayStorage so it's conflicting with any other private function for storage?

contracts/interfaces/IAxelarAmplifierGateway.sol
re1ro: If we do we better do it now. Before releasing the new multisig.

re1ro: Before you were inclined to keep it compatible with the current gateway, so we changed it [back to bytes](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/8c0df6bc62d33c5730d04c2155979ad7ae0a7e72/contracts/governance/BaseWeightedMultisig.sol#L77).

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
-    function _approveContractCall(bytes memory params, bytes32 commandId) internal {
-        (
-            string memory sourceChain,
-            string memory sourceAddress,
-            address contractAddress,
-            bytes32 payloadHash,
-            bytes32 sourceTxHash,
-            uint256 sourceEventIndex
-        ) = abi.decode(params, (string, string, address, bytes32, bytes32, uint256));
+    function _isContractCallApproved(
```
re1ro, line 156: Do you think we need to create internal functions if they are used only in 1 place? Keeping it inline makes contract smaller overall. And doesn't require to scroll up and down when reading the code.

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
 
     /**********************\
     |* External Functions *|
     \**********************/
 
-    function execute(bytes calldata batch) external {
-        (bytes memory data, bytes memory proof) = abi.decode(batch, (bytes, bytes));
+    function execute(SignedCommandBatch calldata signedBatch) external {
+        CommandBatch calldata batch = signedBatch.batch;
+        bytes memory data = abi.encode(batch);
```
re1ro, line 108: Hey maybe we don't need to encode it again but can slice it out from `msg.data`?

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
-
-        (chainId, commandIds, commands, params) = abi.decode(data, (uint256, bytes32[], string[], bytes[]));
+        bool allowOperatorshipTransfer = authModule.validateProof(batchHash, signedBatch.proof);
 
-        if (chainId != block.chainid) revert InvalidChainId();
+        if (chainNameHash != signedBatch.batch.chainNameHash) revert InvalidChainName();
+        if (routerHash != signedBatch.batch.routerHash) revert InvalidRouter();
 
-        uint256 commandsLength = commandIds.length;
+        Command[] memory commands = batch.commands;
```
re1ro, line 120: see if you can make it `calldata` so it doesn't copy commands into memory

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
 
-        if (commandsLength != commands.length || commandsLength != params.length) revert InvalidCommands();
-
-        for (uint256 i; i < commandsLength; ++i) {
-            bytes32 commandId = commandIds[i];
+        for (uint256 i; i < commands.length; ++i) {
+            Command memory command = commands[i];
```
re1ro, line 123: `calldata`?

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
                     continue;
                 }
 
                 allowOperatorshipTransfer = false;
 
-                _transferOperatorship(params[i]);
+                _transferOperatorship(command.params);
             } else {
-                revert InvalidCommand(commandHash);
+                revert InvalidCommand();
```
re1ro, line 142: Before we were just skipping the unknown so the batch can be resubmitted later to process the commands. (Maybe gateway upgrade, etc). Are you sure you want to revert?

contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -9,6 +9,32 @@ import { IAxelarGMPGateway } from './IAxelarGMPGateway.sol';
...
+    \*********/
+
+    enum CommandType {
+        ApproveContractCall,
+        TransferOperatorship
+    }
+
+    struct Command {
+        CommandType commandType;
+        string messageId;
```
re1ro, line 23: Let's make it `bytes` or `bytes32` instead of string. To keep it more efficient as we did with ITS.

And what is it used for?

re1ro: If the length <= 31  we can just store it padded as `bytes32`

contracts/interfaces/IAxelarAmplifierGateway.sol
re1ro: I like `validateMessage`

contracts/interfaces/IAxelarAmplifierGateway.sol
re1ro: Oh I see
```
bytes32 commandId = keccak256(bytes(command.messageId));
```

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
+        valid = _storage().approvals[key];
+
+        if (valid) {
+            delete _storage().approvals[key];
+
+            emit ContractCallExecuted(commandId);
+        }
+    }
+
+    function _approveContractCall(bytes32 commandId, Command memory command) internal {
```
re1ro, line 189: `Command calldata command`?

contracts/interfaces/IAxelarAmplifierGateway.sol
re1ro: We basically repurposed old `payloadHash, sourceTxHash, sourceEventIndex` into the `messageId`. Yeah lets just use `bytes` instead. String encoding (UTF) might mess the binary data

milapsheth: `messageId` is an opaque string in the core protocol. For e.g. it's hex encoded tx hash with event index as string for EVM but can be different for other chains. We want to preserve that. Makes it easier to debug on the explorer as well.

contracts/interfaces/IAxelarAmplifierGateway.sol
milapsheth: it feels a bit weird with `callContract` on the source, but I agree it's simpler

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -63,93 +78,117 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         string calldata sourceAddress,
         bytes32 payloadHash
     ) external override returns (bool valid) {
-        bytes32 key = _getIsContractCallApprovedKey(commandId, sourceChain, sourceAddress, msg.sender, payloadHash);
-        valid = _axelarAmplifierGatewayStorage().approvals[key];
-
-        if (valid) {
-            delete _axelarAmplifierGatewayStorage().approvals[key];
+        valid = _validateContractCall(commandId, sourceChain, sourceAddress, payloadHash);
```
Foivos, line 81: This used to revert, are we sure we want to change the pattern here (`AxelarExecutable` will have to change to account for this as well)

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: it's a private method, so i'm thinking we use `_storage` as a convention for all contracts

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: we don't revert in the existing gateway
https://github.com/axelarnetwork/axelar-cgp-solidity/blob/main/contracts/AxelarGateway.sol#L239

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: it's used twice

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: Amplifier allows re-signing approvals indefinitely unlike core (and relies commandId check to prevent replay), so commands shouldn't be stuck. We also shouldn't get unknown commands. Gateway upgrade is just a GMP approval. Original gateway had more command types with inability to re-sign.

contracts/gateway/AxelarAmplifierGateway.sol
Foivos: Oh I must have had it wrong in my head, this seems good then.

PR #150, opened 2024-04-02
refactor(amplifier)!: split message verification and signer rotation

milapsheth:
Refactors #148 

Split contract call approval and key rotation commands into separate entrypoints. This avoids the complication of dealing with opaque Command type, and isolates key rotation checks, and command id derivation can be handled better.

`verifyMessages` and `rotateSigners` are the respective entrypoints. Proof validation logic is shared.

- [ ] add common `execute` entrypoint for convenience?
- [ ] update tests

### DIFF
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -80,10 +80,6 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
         valid = _validateContractCall(commandId, sourceChain, sourceAddress, payloadHash);
     }
 
-    /***********\
-    |* Getters *|
-    \***********/
-
     function isCommandExecuted(bytes32 commandId) public view override returns (bool) {
         return _storage().commands[commandId];
     }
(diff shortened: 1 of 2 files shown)


contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -14,38 +14,47 @@ interface IAxelarAmplifierGateway is IAxelarGMPGateway {
...
 
-    struct RotateSignersParams {
-        bytes32 nonce;
-        bytes newSigners;
+    struct Proof {
+        // Amplifier domain separator
+        // keccak256(chain_name || amplifier_router_address || axelar_chain_id)
+        bytes32 domainSeparator;
+        // Commitment to the signer set
+        bytes32 signerCommitment;
```
milapsheth, line 34: involves a larger refactor but would be cleaner to move this into the auth module's proof type instead

re1ro: +1

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -92,54 +88,65 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
-
-                isLatestSigners = false;
-
-                _rotateSigners(command.params);
-            } else {
-                revert InvalidCommand();
-            }
+            _approveMessage(commandId, message);
 
             _storage().commands[commandId] = true;
```
re1ro, line 109: Does it Make sense to cache storage slot into a variable? Especially in a loop

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -92,54 +88,65 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
-    function execute(SignedCommandBatch calldata signedBatch) external {
-        CommandBatch calldata batch = signedBatch.batch;
-        bytes memory data = abi.encode(batch);
-
-        // This is the EVM convention for signing non-tx data to domain separate from regular EVM txs.
-        // This should be customized per chain.
-        bytes32 batchHash = ECDSA.toEthSignedMessageHash(keccak256(data));
-
-        bool isLatestSigners = authModule.validateProof(batchHash, signedBatch.proof);
+    function verifyMessages(SignedMessageBatch calldata signedBatch) public {
```
re1ro, line 91: I am still a lil bit confused by the `verifyMessages` method name as we don't actually verify any messages in this context.
How about:
`deliverMessages`
`relayMessages`
`approveMessages` +1
`approveBatch`
`passMessages` hence GMP?

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -92,54 +88,65 @@ contract AxelarAmplifierGateway is IAxelarAmplifierGateway {
...
+
     /**********************\
     |* Internal Functions *|
     \**********************/
 
+    function _verifyProof(CommandType commandType, bytes memory data, Proof calldata proof) internal view returns (bool) {
+        if (domainSeparator != proof.domainSeparator) revert InvalidDomainSeparator();
+
+        // TODO: use SignData struct
+        data = abi.encode(commandType, proof.domainSeparator, proof.signerCommitment, data);
```
re1ro, line 143: Seems odd that we have to double encode data? Could we move to something like?
```
abi.encode(commandType, proof.domainSeparator, proof.signerCommitment, signedRotation.rotation);
```
Also would be nice to have it encoded in the same sequence in the calldata so we could slice it out.

contracts/interfaces/IAxelarAmplifierGateway.sol
```diff
@@ -14,38 +14,47 @@ interface IAxelarAmplifierGateway is IAxelarGMPGateway {
...
+        bytes proof;
     }
 
-    struct Command {
+    struct SignData {
```
re1ro, line 38: SignedData?

re1ro (changes requested):

contracts/gateway/AxelarAmplifierGateway.sol
re1ro: Lets talk and think more about those structs

milapsheth: yeah, haven't settled on that. the encoding needs to be done in amplifier as well. It might be easier to have the last arg as `bytes` to reuse encoding between different command types

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: `verifyMessages` comes from Amplifier gateways. Not settled on it yet

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: it was more expensive when i cached it. I think the compiler optimizes this better due to inlining. I'll do an optimization pass again at the end

contracts/interfaces/IAxelarAmplifierGateway.sol
milapsheth: doesn't include the signatures/proof. but open to other names (this was just a placeholder/isn't being used yet)

PR #154, opened 2024-04-08
feat(amplifier)!: amplifier gateway v2

milapsheth:
[AXE-3660]

- [x] fix tests

[AXE-3660]: https://axelarnetwork.atlassian.net/browse/AXE-3660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Breaking [events](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/154/files#diff-5d120e5bd5652c34710b547eca533fcca2fbb2fc45225a1ea72e1f3f2f3306daR50-R64)
- `ContractCallApproved` now includes the `messageId` instead of tx hash and event index separately
- `SignersRotated` event is used instead of `OperatorshipTransferred` to adhere to the updated naming convention

`commandId` derivation has changed. See `messageToCommandId` [here](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/pull/154/files#diff-7fb6c9a6f0d6e78d4f723798364da22d45c3605755c02a82441aad6f67be0f96R90). It's still guaranteed to be globally unique.


Foivos (commented):
I think one could pass rotate signers into approve messages and vice versa, which is potentially dangerous.

Foivos (approved):

re1ro (approved):

milapsheth:
> I think one could pass rotate signers into approve messages and vice versa, which is potentially dangerous.

Discussed offline, but adding here for completeness. The message hash being signed also depends on `CommandType` which isn't user controlled and avoids this issue.

PR #155, opened 2024-04-09
feat(ValuedExpress): allocating value for token flows

re1ro:
* [x] `AxelarValuedExpressExecutable`: allocating value before `_execute` for token flows


Foivos (commented):
Why did you change the order of the to and from functions? It makes checking for changes quote hard

contracts/express/AxelarValuedExpressExecutable.sol
```diff
@@ -203,31 +204,37 @@ abstract contract AxelarValuedExpressExecutable is ExpressExecutorTracker, IAxel
...
         address expressExecutor,
         address tokenAddress,
         uint256 value
-    ) internal {
+    ) internal virtual {
         if (value == 0) return;
 
         if (tokenAddress == address(0)) {
-            payable(expressExecutor).safeNativeTransfer(value);
+            if (msg.value < value) revert InsufficientValue();
```
StephenFluin, line 221: Should the value be exact?

re1ro: What if there is more value provided to make another GMP call and pay gas? I think we should allow for extra value to be passed.

re1ro:
@Foivos yeah sorry, the only real change there is to allow to pass more value than estimated. To pay gas for consequent GMP calls etc

```
if (msg.value < value) revert InsufficientValue();
```

StephenFluin (dismissed):

PR #176, opened 2024-06-21
chore: add contractId in AxelarAmplifierGatewayProxy

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4196

https://github.com/axelarnetwork/axelar-contract-deployments/pull/267#discussion_r1647286089

### DIFF
--- a/contracts/gateway/AxelarAmplifierGatewayProxy.sol
+++ b/contracts/gateway/AxelarAmplifierGatewayProxy.sol
@@ -5,6 +5,8 @@ pragma solidity ^0.8.0;
 import { Proxy } from '../upgradable/Proxy.sol';
 
 contract AxelarAmplifierGatewayProxy is Proxy {
+    bytes32 private constant CONTRACT_ID = keccak256('axelar-amplifier-gateway-proxy');
+
     constructor(
         address implementationAddress,
         address owner,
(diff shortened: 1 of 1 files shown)


contracts/gateway/AxelarAmplifierGatewayProxy.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity ^0.8.0;
 import { Proxy } from '../upgradable/Proxy.sol';
 
 contract AxelarAmplifierGatewayProxy is Proxy {
+    bytes32 private constant CONTRACT_ID = keccak256('axelar-amplifier-gateway-proxy');
```
milapsheth, line 8:
```suggestion
    bytes32 public constant CONTRACT_ID = keccak256('axelar-amplifier-gateway');
```
this is confusing since there's already a method called `contractId` with a different id. just make it public instead and use it internally to avoid duplication

contracts/gateway/AxelarAmplifierGatewayProxy.sol
```diff
@@ -14,4 +16,8 @@ contract AxelarAmplifierGatewayProxy is Proxy {
     function contractId() internal pure override returns (bytes32) {
         return keccak256('axelar-amplifier-gateway');
     }
+
+    function getContractId() public pure returns (bytes32) {
+        return CONTRACT_ID;
+    }
```
milapsheth, line 22:
```suggestion
        return CONTRACT_ID;
    }
```
contractId is already exposed via the implementation, so exposing a different value via another getter can be an issue

milapsheth (approved):

PR #177, opened 2024-06-26
fix(governance): add check for strict Multisig threshold validation

blockchainguyy:
# Strict Weighted Multisig threshold validation
This PR adds a check to revert if there are any leftover signatures, when threshold is crossed to make the threshold validation strict.

- [x] add error and check in smart contract for strict validation.
- [x] remove positive test for signatures exceeding required signatures.
- [x] add negative test to revert in case of signatures exceeding required signatures.
- [x] add comment stating operator is an optional param in setup, which can be removed in the future
- [x] remove check for transferring operator role to address 0.

[Jira Ticket](https://axelarnetwork.atlassian.net/browse/AXE-4308)

### DIFF
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -53,7 +53,7 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
     function _setup(bytes calldata data) internal override {
         (address operator_, WeightedSigners[] memory signers) = abi.decode(data, (address, WeightedSigners[]));
 
-        // operator is an optional parameter. The gateway owner can set it later if needed.
+        // operator is an optional parameter. The gateway owner can set it later via `transferOperatorship` if needed.
         // This also simplifies setup for upgrades so if the current operator doesn't need to be changed, then it can be skipped, instead of having to specify the current operator again.
         if (operator_ != address(0)) {
             _transferOperatorship(operator_);
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -126,6 +126,7 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
 
     /**
      * @notice Transfer the operatorship to a new address.
+     * @dev The owner or current operator can set the operator to address 0.
      * @param newOperator The address of the new operator.
      */
     function transferOperatorship(address newOperator) external onlyOperatorOrOwner {
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -53,6 +53,7 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
     function _setup(bytes calldata data) internal override {
         (address operator_, WeightedSigners[] memory signers) = abi.decode(data, (address, WeightedSigners[]));
 
+        // operator is an optional parameter, it can also be removed in the future to make ownership completely decentralized
         if (operator_ != address(0)) {
             _transferOperatorship(operator_);
         }
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -129,8 +129,6 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
      * @param newOperator The address of the new operator.
      */
     function transferOperatorship(address newOperator) external onlyOperatorOrOwner {
-        if (newOperator == address(0)) revert InvalidOperator();
-
         _transferOperatorship(newOperator);
     }
 
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -53,7 +53,8 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
     function _setup(bytes calldata data) internal override {
         (address operator_, WeightedSigners[] memory signers) = abi.decode(data, (address, WeightedSigners[]));
 
-        // operator is an optional parameter, it can also be removed in the future to make ownership completely decentralized
+        // operator is an optional parameter. The gateway owner can set it later if needed.
+        // This also simplifies setup for upgrades so if the current operator doesn't need to be changed, then it can be skipped, instead of having to specify the current operator again.
         if (operator_ != address(0)) {
             _transferOperatorship(operator_);
         }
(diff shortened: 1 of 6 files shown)


milapsheth:
Rename PR to `fix(governance): ..` since it's a non breaking fix in the governance contract folder

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -216,8 +216,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             // accumulating signatures weight
             totalWeight = totalWeight + signers[signerIndex].weight;
 
-            // weight needs to reach or surpass threshold
-            if (totalWeight >= weightedSigners.threshold) return;
+            // weight needs to reach threshold
+            if (totalWeight >= weightedSigners.threshold) {
+                // check if number of signatures exceed required amount of signatures
+                if (signaturesLength - 1 != i) revert InvalidSignaturesLength(signaturesLength, i + 1);
```
milapsheth, line 222:
```suggestion
                if (i + 1 != signaturesLength) revert InvalidSignaturesLength(i + 1, signaturesLength);
```
More common to keep i first

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -216,8 +216,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             // accumulating signatures weight
             totalWeight = totalWeight + signers[signerIndex].weight;
 
-            // weight needs to reach or surpass threshold
-            if (totalWeight >= weightedSigners.threshold) return;
+            // weight needs to reach threshold
+            if (totalWeight >= weightedSigners.threshold) {
+                // check if number of signatures exceed required amount of signatures
```
milapsheth, line 221:
```suggestion
                // check if there are redundant signatures
```
Perhaps a bit more clear

contracts/interfaces/IBaseWeightedMultisig.sol
```diff
@@ -9,6 +9,7 @@ interface IBaseWeightedMultisig {
     error LowSignaturesWeight();
     error InvalidWeights();
     error DuplicateSigners(bytes32 signersHash);
+    error InvalidSignaturesLength(uint256 provided, uint256 required);
```
milapsheth, line 12:
```suggestion
    error InvalidSignaturesLength(uint256 provided, uint256 required);
```

```suggestion
    error InvalidSignaturesLength(uint256 required, uint256 provided);
```

contracts/governance/BaseWeightedMultisig.sol
milapsheth, line 222: Update the validate Proof docstring to mention this requirement as well

test/governance/BaseWeightedMultisig.js
```diff
@@ -481,6 +473,18 @@ describe('BaseWeightedMultisig', () => {
                     'LowSignaturesWeight',
                 );
             });
+
+            it('reject the proof if number of total signatures exceed required signatures', async () => {
```
milapsheth, line 477:
```suggestion
            it('reject the proof if there are redundant signatures', async () => {
```

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -53,6 +53,7 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
     function _setup(bytes calldata data) internal override {
         (address operator_, WeightedSigners[] memory signers) = abi.decode(data, (address, WeightedSigners[]));
 
+        // operator is an optional parameter, it can also be removed in the future to make ownership completely decentralized
```
milapsheth, line 56:
```suggestion
        // operator is an optional parameter. The gateway owner can set it later if needed.
        // This also simplifies setup for upgrades so if the current operator doesn't need to be changed, then it can be skipped, instead of having to specify the current operator again.
```
whether ownership is decentralized depends on what the actual owner is set to. The contract code doesn't know about this, so any comments should be with respect to the roles in the contract

milapsheth, line 56: the comment about the owner being able to remove / transfer operator to address(0), should be made on the `transferOperatorship` method

milapsheth: add a unit test for this if there isn't one

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -107,6 +107,7 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
      * @param proof The multisig proof data
      * @return isLatestSigners True if the proof is from the latest signer set
      * @dev The proof data should have signers, weights, threshold and signatures encoded
+     *      The proof is only valid if signers weight crosses threshold and there are no redundant signatures
```
milapsheth, line 110:
```suggestion
     *      The proof is only valid if the signers weight crosses the threshold and there are no redundant signatures
```

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -216,8 +217,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             // accumulating signatures weight
             totalWeight = totalWeight + signers[signerIndex].weight;
 
-            // weight needs to reach or surpass threshold
-            if (totalWeight >= weightedSigners.threshold) return;
+            // weight needs to reach threshold
+            if (totalWeight >= weightedSigners.threshold) {
+                // check if there are redundant signatures
+                if (signaturesLength - i == 1) return;
```
milapsheth, line 223:
```suggestion
                if (i + 1 == signaturesLength) return;
```
feels less idiomatic to check it like that

contracts/governance/BaseWeightedMultisig.sol
```diff
@@ -216,8 +217,13 @@ abstract contract BaseWeightedMultisig is IBaseWeightedMultisig {
             // accumulating signatures weight
             totalWeight = totalWeight + signers[signerIndex].weight;
 
-            // weight needs to reach or surpass threshold
-            if (totalWeight >= weightedSigners.threshold) return;
+            // weight needs to reach threshold
+            if (totalWeight >= weightedSigners.threshold) {
+                // check if there are redundant signatures
```
milapsheth, line 222:
```suggestion
                // validate the proof if there are no redundant signatures
```

contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -53,6 +53,8 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
     function _setup(bytes calldata data) internal override {
         (address operator_, WeightedSigners[] memory signers) = abi.decode(data, (address, WeightedSigners[]));
 
+        // operator is an optional parameter. The gateway owner can set it later if needed.
```
milapsheth, line 56:
```suggestion
        // operator is an optional parameter. The gateway owner can set it later via `transferOperatorship` if needed.
```

milapsheth (approved):

PR #186, opened 2024-09-10
feat: create gmp tracker event interface

SGiaccobasso:
https://axelarnetwork.atlassian.net/browse/AXE-5550


contracts/interfaces/IAxelarGMPTracker.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+interface IAxelarGMPTracker {
+    event GMPTracker(
```
npty, line 6: perhaps rename it to `GMPVolume`? `emit GMPVolume(...)` sounds a bit more natural to me.

ecdsafu (commented):
1. For our RFQ, sender won't be on the chain that this event is being emitted, the sender of the transaction being emitted will just be a relayer for our settlement process, so this should be a string. Could be a sender from Sui, Sol, or Cosmos. 
3. We don't have decimals available, so wouldn't be able to emit that. 
4. Swaps have a fromToken and toToken, so a bridge amount and single tokenAddress is less relevant here. Better to have fromToken, fromAmount, toToken and toAmount for us. We could arbitrarily pick the destination token if you want just one.

canhtrinh:
> 1. For our RFQ, sender won't be on the chain that this event is being emitted, the sender of the transaction being emitted will just be a relayer for our settlement process, so this should be a string. Could be a sender from Sui, Sol, or Cosmos.
=> Thanks, yeah we'll change to a string

> 2. We don't have decimals available, so wouldn't be able to emit that.
=> How do you represent amounts then?

> 3. Swaps have a fromToken and toToken, so a bridge amount and single tokenAddress is less relevant here. Better to have fromToken, fromAmount, toToken and toAmount for us. We could arbitrarily pick the destination token if you want just one.
=> Hmm, yeah we just want one. I'll want some team buy-in, but perhaps we can accept both `from` and `to` so you don't have to pick

ecdsafu:
> > 1. For our RFQ, sender won't be on the chain that this event is being emitted, the sender of the transaction being emitted will just be a relayer for our settlement process, so this should be a string. Could be a sender from Sui, Sol, or Cosmos.
> 
> => Thanks, yeah we'll change to a string
> 
> 
> 
> > 2. We don't have decimals available, so wouldn't be able to emit that.
> 
> => How do you represent amounts then?
> 
> 
> 
> > 3. Swaps have a fromToken and toToken, so a bridge amount and single tokenAddress is less relevant here. Better to have fromToken, fromAmount, toToken and toAmount for us. We could arbitrarily pick the destination token if you want just one.
> 
> => Hmm, yeah we just want one. I'll want some team buy-in, but perhaps we can accept both `from` and `to` so you don't have to pick

1. Sounds good, thanks 

2. We have off chain mappings for the token/chain/decimals and amount is always represented in the decimals of that token/chain

3. Ok sounds good, if you just want one we can choose only one, will leave it with you if you want to add the second token or not

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title IAxelarGMPTransfer
+ * @dev Interface for tracking asset value transfers using General Message Passing (GMP) calls in the Axelar Network.
+ * This interface defines an event that should be emitted when a GMP transfer occurs,
+ * allowing for standardized volume tracking across different implementations.
+ */
+interface IAxelarGMPTransfer {
```
milapsheth, line 11: update a test contract to make use of this on send and execute for coverage, e.g. `GMPExecutableTest`. The corresponding [test](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/febcd9fb2b4c0a6db23f5265f3ae9e115a3f2fc6/test/executable/GMPExecutable.js#L109-L110) for the event can be updated as well

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+     */
+    event AxelarGMPTransfer(
+        address indexed sender,
+        string sourceChain,
+        string destinationChain,
+        string destinationAddress,
+        string recipientAddress,
+        string tokenAddress,
+        uint256 amount,
+        uint256 decimals
```
milapsheth, line 31: Don't think we need decimals. Axelarscan can read from the config, or query it off-chain. The app would need to query it on-chain otherwise which is unnecessarily expensive

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+     * @param tokenAddress The address of the token contract on the source chain.
+     * @param amount The amount (in atomic units) of tokens transferred.
+     * @param decimals The number of decimal places for the token.
+     */
+    event AxelarGMPTransfer(
+        address indexed sender,
+        string sourceChain,
+        string destinationChain,
+        string destinationAddress,
+        string recipientAddress,
```
milapsheth, line 28:
```suggestion
        bytes recipient,
```
No guarantees that apps use strings for the underlying recipient (they often won't since it's inefficient, e.g. ITS)

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+     * @param sender The address of the caller that initiated the transfer on the source chain.
+     * @param sourceChain The Axelar chain identifier of the source chain.
+     * @param destinationChain The Axelar chain identifier of the destination chain.
+     * @param destinationAddress The address of the contract on the destination chain that receives the transfer.
+     * @param recipientAddress The address of the final recipient of the transferred assets on the destination chain.
+     * @param tokenAddress The address of the token contract on the source chain.
+     * @param amount The amount (in atomic units) of tokens transferred.
+     * @param decimals The number of decimal places for the token.
+     */
+    event AxelarGMPTransfer(
```
milapsheth, line 23: The same event shouldn't be used for both source and destination.

On the source side, `sourceChain` and `sender` are obvious from the event. `sourceChain` identifier isn't actually available to apps even (it's an identifier at the protocol level, we could add it to the gateway but it'll still be costly to query it).

On the destination side, `destinationChain` isn't available similarly, and `destinationAddress` isn't available as a string (costly to convert) but also not needed, these values are clear from the event itself. `string sourceAddress` should be used to mimic GMP. `bytes sender` could be added as well for the true sender in case it's different than the source contract address (e.g. ITS transfer)

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+        address indexed sender,
+        string sourceChain,
+        string destinationChain,
+        string destinationAddress,
```
milapsheth, line 27:
```suggestion
        string destinationContractAddress,
```

contracts/interfaces/IAxelarGMPTransfer.sol
milapsheth, line 23: Axelarscan should only trust this event from whitelisted apps, otherwise anyone can deploy a GMP contract and emit events with invalid value causing indexing issues (e.g. $1T transfer amount)

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,33 @@
...
+        string recipientAddress,
+        string tokenAddress,
```
milapsheth, line 29:
```suggestion
        address token,
```
This is a token on the EVM chain the event is being emitted on. It should always be an `address`.

milapsheth (changes requested):

contracts/interfaces/IAxelarGMPTransfer.sol
milapsheth: While I'd prefer to require the event to only be emitted on the destination chain, I suppose it's convenient to allow source to emit it as well, if the destination isn't a contract it controls, or if the GMP call wasn't executed/takes a while due to finality.

milapsheth, line 23: check with devrel as well on naming thoughts. some other options `AxelarMessageValue`, `AxelarTransferValue`, `AxelarGMPValue`

canhtrinh: I was wondering if we should have different events depending on whether it was being emitted from the source chain or the executable on the destination chain. So yeah, we can consider creating two different events, one for src, one for dest?

contracts/interfaces/IAxelarGMPTransfer.sol
canhtrinh: Yeah, we should do that. @SGiaccobasso , can you run with this?

contracts/interfaces/IAxelarGMPTransfer.sol
canhtrinh: Interesting, okay. @SGiaccobasso - let's remove decimals

contracts/interfaces/IAxelarGMPTransfer.sol
milapsheth: Yep, should have different events with only the applicable fields in each case

milapsheth: Another option is that axelarscan can label it in the GMP view for a non whitelisted app but not include it in indexing (such as volume, sort by value etc.) until the contract address is whitelisted (e.g. given a label/name in the axelarscan config)

milapsheth: Other naming suggestions

`AxelarTransferSent` and `AxelarTransferReceived`

There's also InterchainTransferSent and Received, similar to what ITS does. I was considering using these events in ITS itself as well, although there're some other metadata that ITS events have

contracts/interfaces/IAxelarGMPTransfer.sol
pjenviri: IMO, from axelarscan end, I prefer to have `symbol` and `decimals` in place if possible. If not, it remains viable, but additional queries will be required.

ecdsafu:
Looking good, for our RFQ we will always emit from source on the hub chain. If you add decimals back in, we won't be able to support anyway and would leave at 0

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,53 @@
...
+     * @param destinationContractAddress The address of the contract on the destination chain that receives the transfer.
+     * @param recipient The address of the final recipient of the transferred assets on the destination chain.
+     * @param token The address of the token contract on the source chain.
+     * @param amount The amount (in atomic units) of tokens transferred.
+     */
+    event InterchainTransferSent(
+        string destinationChain,
+        string destinationContractAddress,
+        bytes recipient,
+        address token,
```
milapsheth, line 24:
```suggestion
        address indexed token,
```

contracts/interfaces/IAxelarGMPTransfer.sol
milapsheth, line 1: Name this `IInterchainTransfer.sol` perhaps?

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,53 @@
...
+     * @param sender The address of the sender in case it is different from the source contract address
+     * @param recipient The address of the final recipient of the transferred assets on the destination chain.
+     * @param token The address of the token contract on the destination chain.
+     * @param amount The amount (in atomic units) of tokens received.
+     */
+    event InterchainTransferReceived(
+        string sourceChain,
+        string sourceAddress,
+        bytes sender,
+        bytes recipient,
```
milapsheth, line 49:
```suggestion
        address indexed recipient,
```
This is an address on the current chain, so it can use the native type

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,53 @@
...
+        address token,
```
milapsheth, line 50:
```suggestion
        address indexed token,
```

contracts/interfaces/IAxelarGMPTransfer.sol
```diff
@@ -0,0 +1,53 @@
...
+     * @param destinationChain The Axelar chain identifier of the destination chain.
+     * @param destinationContractAddress The address of the contract on the destination chain that receives the transfer.
+     * @param recipient The address of the final recipient of the transferred assets on the destination chain.
+     * @param token The address of the token contract on the source chain.
+     * @param amount The amount (in atomic units) of tokens transferred.
+     */
+    event InterchainTransferSent(
+        string destinationChain,
+        string destinationContractAddress,
+        bytes recipient,
```
milapsheth, line 23:
```suggestion
        address indexed sender,
        bytes recipient,
```

contracts/test/executable/GMPExecutableWithTokenTest.sol
```diff
@@ -3,32 +3,40 @@
...
+        string calldata sourceChain,
+        string calldata sourceAddress,
         bytes calldata payload,
         string calldata tokenSymbol,
         uint256 amount
     ) internal override {
-        uint256 num = abi.decode(payload, (uint256));
-        emit ReceivedWithToken(num, gatewayWithToken().tokenAddresses(tokenSymbol), amount);
+        // Emit InterchainTransferReceived event
+        emit InterchainTransferReceived(
```
milapsheth, line 33: Technically, this is meant to be used for GMP execute, instead of GMP with token where the value is obvious. But we can help clean this up for you with an example in another PR

contracts/test/executable/GMPExecutableWithTokenTest.sol
```diff
@@ -3,32 +3,40 @@
...
+            sourceChain,
+            sourceAddress,
+            payload,
```
milapsheth, line 36:
```suggestion
            bytes(sourceAddress),
```
payload doesn't make sense here

milapsheth (approved):

ecdsafu:
Hey guys, are these final and good to go into our mainnet contracts? We're deploying soon

ecdsafu:
Hey guys, just thought of an issue with this approach. Since all events will be emitted from our hub chain, there is no way to express the fromChain for the order if we use the InterchainTokenSent event. You will have the toChain in the event, but not know what the fromChain is.  

Would also be helpful to know what UI this is going to be displayed on. Will each event be on axelarscan as a bridge tx with a fromChain toChain and amount?

milapsheth (approved):

PR #189, opened 2024-09-24
chore: use named return

ahramy:
[AXE-5205](https://axelarnetwork.atlassian.net/browse/AXE-5205)
- Use the named return variables for efficient gas and consistency.
- Updated comments accordingly.

[AXE-5205]: https://axelarnetwork.atlassian.net/browse/AXE-5205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -118,7 +118,7 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
 
     /**
      * @notice Returns the address of the gateway operator.
-     * @return address The address of the operator.
+     * @return The address of the operator.
      */
     function operator() external view returns (address) {
         return _axelarAmplifierGatewayStorage().operator;
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -118,10 +118,10 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
 
     /**
      * @notice Returns the address of the gateway operator.
-     * @return operatorAddress The address of the operator.
+     * @return address The address of the operator.
      */
-    function operator() external view returns (address operatorAddress) {
-        operatorAddress = _axelarAmplifierGatewayStorage().operator;
+    function operator() external view returns (address) {
+        return _axelarAmplifierGatewayStorage().operator;
     }
 
     /**
--- a/contracts/gateway/AxelarAmplifierGateway.sol
+++ b/contracts/gateway/AxelarAmplifierGateway.sol
@@ -113,15 +113,15 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
      * @return isLatestSigners True if provided signers are the current ones
      */
     function validateProof(bytes32 dataHash, Proof calldata proof) external view returns (bool isLatestSigners) {
-        return _validateProof(dataHash, proof);
+        isLatestSigners = _validateProof(dataHash, proof);
     }
 
     /**
      * @notice Returns the address of the gateway operator.
-     * @return The address of the operator.
+     * @return operatorAddress The address of the operator.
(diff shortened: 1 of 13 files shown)


contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -113,15 +113,15 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
...
      */
     function validateProof(bytes32 dataHash, Proof calldata proof) external view returns (bool isLatestSigners) {
-        return _validateProof(dataHash, proof);
+        isLatestSigners = _validateProof(dataHash, proof);
     }
 
     /**
      * @notice Returns the address of the gateway operator.
-     * @return The address of the operator.
+     * @return operatorAddress The address of the operator.
```
milapsheth, line 121: don't think the naming is very useful here. it's implied by the function name. we should use named variables only for non-obvious cases

contracts/gateway/BaseAmplifierGateway.sol
```diff
@@ -224,15 +229,16 @@ abstract contract BaseAmplifierGateway is IBaseAmplifierGateway {
...
     function _messageHash(
         bytes32 commandId,
         string calldata sourceChain,
         string calldata sourceAddress,
         address contractAddress,
         bytes32 payloadHash
-    ) internal pure returns (bytes32) {
-        return keccak256(abi.encode(commandId, sourceChain, sourceAddress, contractAddress, payloadHash));
+    ) internal pure returns (bytes32 hash) {
+        hash = keccak256(abi.encode(commandId, sourceChain, sourceAddress, contractAddress, payloadHash));
```
milapsheth, line 241: same, doesn't need to be named. although good to have the natspec doc

contracts/governance/BaseMultisig.sol
```diff
@@ -53,18 +53,18 @@ contract BaseMultisig is IBaseMultisig {
 
     /**
      * @notice Returns the current signer threshold
-     * @return uint The signer threshold
+     * @return threshold The signer threshold
      */
-    function signerThreshold() external view override returns (uint256) {
-        return signers.threshold;
+    function signerThreshold() external view override returns (uint256 threshold) {
```
milapsheth, line 58: same, avoid all these obvious cases

contracts/gateway/AxelarAmplifierGateway.sol
milapsheth: this is another example of not taking the report at face value. we should only make the change that actually makes sense to us

ahramy: the report says modify only one line for `axelar-gmp-sdk-solidity`, but I reviewed all files and thought it's better to be consistent with the pattern (using named return). I agree with that `we should use named variables only for non-obvious cases`. Reviewed all files and updated.

contracts/gateway/BaseAmplifierGateway.sol
ahramy: updated.

contracts/governance/BaseMultisig.sol
ahramy: updated.

contracts/governance/InterchainGovernance.sol
```diff
@@ -174,7 +174,8 @@ contract InterchainGovernance is AxelarGMPExecutable, TimeLock, Caller, IInterch
     }
 
     /**
-     * @dev Get proposal hash using the target, callData, and nativeValue
+     * @notice Get proposal hash using the target, callData, and nativeValue
+     * @return The proposal hash
```
milapsheth, line 178: This is another contract fixed in production. Prefer to avoid changing it unless needed

milapsheth (approved):

contracts/governance/InterchainGovernance.sol
ahramy: Got it. Reverted change.

PR #227, opened 2026-05-26
feat(gateway): pauseable gateway

MakisChristou:


contracts/gateway/AxelarAmplifierGateway.sol
```diff
@@ -113,6 +119,22 @@ contract AxelarAmplifierGateway is BaseAmplifierGateway, BaseWeightedMultisig, U
...
 
+    /**
+     * @notice Pauses or unpauses the gateway. Callable by the operator (emergency EOA)
+     * or the owner (governance contract or EOA).
+     * @dev Pause freezes `callContract` and gates `validateMessage` / `validateContractCall`
+     * to the owner only; `approveMessages` and `rotateSigners` stay open. Admin functions
+     * remain callable so governance can always recover.
+     * @param isPaused True to pause, false to unpause.
+     */
+    function setPauseStatus(bool isPaused) external onlyOperatorOrOwner {
```
jakovmitrovski, line 130: NIT: I'd leave these as 2 separate functions `pause` that calls `_pause` and `unpause` that calls `_unpause`, but maybe it is just a personal preference.

jakovmitrovski (approved):

contracts/gateway/AxelarAmplifierGateway.sol
MakisChristou: I just wanted to be consistent with how ITS does it.

jakovmitrovski: 👍

dfapel (approved):

